PR #40, opened 2021-11-06
Attempt to automate containerization

kwalcock:


kwalcock:
This is to some extent practice for other containers, like eidos, ConceptAligner, wmexchanger, etc.

PR #42, opened 2021-11-16
Belief reader

mithunpaul08:
will need some help/advice from @kwalcock before this merge on reorganization of code
 
1. i moved the TSVPrinter and JSONPrinter under utils. However, that was more for utility purposes since i wanted it to be accessible in BeliefReader and VariableReader. But I am not sure if that is a final good place to be.
2. Inside TSVPrinter, i created a new outputBeliefMentions to write to tsv- which needs to be merged with outputMentions or have one single function to write both Variables and Beliefs. I kept it separate because a)i wanted to replace Assignment with Belief in the label matching "_.label matches "Belief". that can be easily passed from the parent function, however. b) outputMentions has lots of context writing code while beliefs don't have context. at least for now.
3. JSONPrinter- doesn't have outputBeliefMentions yet. But that is only because I wanted us to decide on no.2 first.


MihaiSurdeanu:
@kwalcock : can you please take a look at @mithunpaul08 's question about the Printers?
Once we merge this, i'd like to reorganize the code a bit. In particular, the "beliefs" should be under "org.clulab.habitus". I can do this once we merge this PR.

kwalcock:
One tiny thing I'd change is the unicode characters in TestBeliefReader.scala in sent15.  Why one quote was escaped but not the other one was a mystery.  More importantly, someone's editor is bound to eventually mess them up.  Maybe it's just policy.  

BTW, that's the test that is failing anyway.  2 mentions are being found.

Somehow I let a small number of semicolons get into the code and now they've been copied.  Could @mithunpaul08 remove them before they proliferate any further?  Search and replace should suffice.  It's my fault.  Thanks!

I'm not sure about adding documents to the `in` directory.  That's not usually done in the repo.

kwalcock:
The move of the Printers to utils is perfect.  They could have been accessed where they were, but then Belief would have needed to know something about Variables, and that's not good.

Instead of mixing Belief and Variables printing function in the utility class, I would be inclined to make it a base class containing any shared functionality and then make a VariablesJsonPrinter and BeliefJsonPrinter etc. that know about the defails, for example what to filter on and whether any context stuff is involved.  For the most part this would be cutting and pasting some code, but it would keep things separate and with little duplication.

MihaiSurdeanu:
I completely agree. @mithunpaul08 : can you please take a shot at this Printer subclasses? One for vars, one for beliefs.

mithunpaul08:
ok. i might be able to get to this only tomorrow. I have to prepare for tomorrow's lecture first.

mithunpaul08:
all the todos have been addressed. please merge.

MihaiSurdeanu:
Merged. Thank you @mithunpaul08 and @kwalcock !

PR #50, opened 2021-11-22
bug fix. : was printing 2 lines when 2 contexts were present in the same sentences

mithunpaul08:


MihaiSurdeanu:
Thanks @mithunpaul08 ! But I still don't understand what the problem was, and what the solution does.
Can you please explain it here?

kwalcock:
Thanks for adding the spaces.  The general idea is to follow https://docs.scala-lang.org/style/.

mithunpaul08:
The problem was caused because there were no context words present in the sentence. My code was checking if context.isempty print just the original 3 parts of the mention(trigger,, mention and the sentence ); if not empty, print all the context details. However, for the case with no context words, I wasn't sending an Empty data structure, I was sending one with N/A in it. I added a check to send Empty as context.

mithunpaul08:
sorry I accidentally closed the PR instead of clicking comment...please merge

mithunpaul08:
actually never mind. i just found an edge case where the fix doesn't work..hold on..wil.l reopen PR with fix

kwalcock:
Heads up, testing on Travis is down.  We ran out of credits.  I will move testing somewhere else, perhaps github.

MihaiSurdeanu:
Merged. Thanks @mithunpaul08 !

src/main/scala/org/clulab/habitus/utils/TsvPrinter.scala
```diff
@@ -70,27 +70,27 @@ class TsvPrinter(outputFilename: String) {
...
+                    contexts(i)(0).mostFreqDate
                   }\t${
-                    context.mostFreqCrop0Sent
+                    contexts(i)(0).mostFreqCrop0Sent
                   }\t${
-                    context.mostFreqCrop1Sent
+                    contexts(i)(0).mostFreqCrop1Sent
                   }\t${
-                    context.mostFreqCrop
+                    contexts(i)(0).mostFreqCrop
```
kwalcock, line 91: I'm concerned about this (0).  If only that one is used, why is there room for more?  Before there was a loop through potentially multiple contexts.

kwalcock (commented):
I'm not really sure what is in these contexts, but this looks odd.

src/main/scala/org/clulab/habitus/utils/TsvPrinter.scala
MihaiSurdeanu: Fair point. @mithunpaul08 : can you please give an example of a sentence with multiple contexts?

mithunpaul08: '''Farmers’ sowing dates ranged from 14 to 31 July for the WS and from 3 to 11 March for the DS.'''

ODIN extracts two events for this sentence (one each for each date range) , but for the same trigger (sowing dates). That is expected right?

Anyway, coming back to the point, the contexts(0) thing is vestigial from the bug fix. Rather, what was happening was that, if there are multiple events in the same sentence, i was creating a separate context object for each mention. Thinking of it now, it is not necessary because all the context counts like LOC,CROP etc (within n sentences away) should remain same wrt to multiple events as long as they occur in the same sentence. . Mihai, your email mentions this is the correct assumption.

Anyway, the original duplication bug was caused because of a for loop (me running through the contexts) for each sorted mentions(which already had a list of multiple mentions). Let me go back create a new branch and make it just one context for all mentions in the same sentence. Functionality will remain the same. I am also in the middle of cleaning up all those long list of contexts(i).mostfreq* with more simple data structures.

MihaiSurdeanu:
Cool. Thanks!

PR #54, opened 2021-11-23
Mithun fertilizer context

mithunpaul08:


MihaiSurdeanu:
Thanks @mithunpaul08 !
This makes sense, but there are conflicts with main. Can you please solve them?

mithunpaul08:
fixed conflicts. actually just hold on with this merge. i want to add a few more test cases. And i want to clean up the printer thing. working with keith on this. will let you know soon.

mithunpaul08:
fixed json printing also.  @kwalcock  do you want to close the other [PR](https://github.com/clulab/habitus/pull/56) @MihaiSurdeanu  if keith says ok can you merge this PR?

kwalcock:
No, the other shouldn't be closed.  The json one here will, for example, still loop over the possible ArrayBuffer elements and create an array of objects, even if there is only ever one.  The tsv version has just used head instead of (0), but there is still an ArrayBuffer.  If head is the proper solution, then I should add it to the other branch and merge with yours.

MihaiSurdeanu:
Yep. Thanks!

kwalcock:
If contexts.nonEmpty but !contexts.contains(i), the tsv version will not print anything.  Is that intended?

mithunpaul08:
@kwalcock you are right. i don't need an array buffer itself since ContextDetails(i) will always contain only one context now. i have changed the context extractor code and TSV printer to reflect that. However, I am not able to find the right syntax for the jobjects thing in jsonPrinter. Will you be able to have a look at it?

kwalcock:
This has been superseded by #58 and the branch [mithun_fertiziler_context_2](https://github.com/clulab/habitus/tree/mithun_fertilizer_context_2).  Please use that branch instead.  It can be renamed to match this branch's name if this one is first deleted.

Issue #46, opened 2021-11-18
Change JSON output to JSON lines (jsonl)


kwalcock:
With one JSON statement per line, output files can be more easily concatenated, as easily as headerless CSV files.

PR #56, opened 2021-11-23
Match TSV output

kwalcock:


kwalcock:
This is slightly different than the TSV version because for the
```scala
contextDetailsMap: mutable.Map[Int, ArrayBuffer[ContextDetails]]
```
this prints out all values in the ArrayBuffer while the TSV version prints only the head (0).  I'm waiting for clarification on that.

Also, in #46 I pondered changing to jsonl format.  I'm not sure what the constraints are.  It is not implemented here, but if the format is changing anyway, now would be a good time to do it.

MihaiSurdeanu:
What is jsonl?

kwalcock:
Json lines.  One json value per line.  https://jsonlines.org/.  This would more closely mimic the TSV output.  I'm otherwise wrapping the entire output in an array, which is also OK.  Without the array wrapping, one can just concatenate output files, though.

MihaiSurdeanu:
Thanks!
Yes, this seems like a good fit.

PR #60, opened 2021-11-25
Add jsonl output

kwalcock:
Experiment with an autocloser of multiple similar resources.  Scala 2.13 has library functions for something similar in Using.Manager.


kwalcock:
This adds jsonl output and also the first version of a library to close multiple resources.  If it works here, it might be promoted to eidos or processors.

kwalcock:
This will soon need to be updated with some accounting for N/A that was mentioned in other threads.  I hope to see this merged so that all those changes can happen at the same time.

MihaiSurdeanu:
Very nice. Thanks @kwalcock !

PR #61, opened 2021-11-29
Fix copulative pattern

hubert10:
Hi @MihaiSurdeanu,

Kindly review this fix on copulative-pattern.

Hubert.


MihaiSurdeanu:
Thanks @hubert10 !

I think this rule:
@variable:Variable (","? @value:Value)+ ","? ("and" @value:Value)?
forces an enumeration to have at least 3 elements, which does not cover the situation with just 2 values with "and" in between. I propose this variant instead:
@variable:Variable (("," | "and")? @value:Value)+ 

Also, can you please add a unit test for that example you had?
Thanks!

hubert10:
Hi @MihaiSurdeanu <https://github.com/MihaiSurdeanu>,

Thanks for the suggestion. I made an update, however your rule does not
cover where the last item of the enumeration is like *, and* ..i.e *'comma'*
followed by '*and*' for example: "*In the SRV, farmers plant Sahel 108,
Sahel 150, Sahel 154, Sahel 134, and Nerica*."
So your rule can perfectly capture patterns such as "*In the SRV, farmers
plant Sahel 108, Sahel 150, Sahel 154, Sahel 134 and Nerica."*

For the example I shared with you before, I added it here
https://github.com/clulab/habitus/blob/fix-copulative-pattern/src/test/scala/org/clulab/habitus/variables/TestVariableReader.scala#L573

Thanks for reviewing this PR.

Hubert.

MihaiSurdeanu:
Good catch @hubert10 !
How about changing the rule to:
@variable:Variable (("," | "and")* @value:Value)+

This allows 0 or more of commas and "and" to happen. Previously, it was just 0 or 1.

hubert10:
Hi @MihaiSurdeanu <https://github.com/MihaiSurdeanu>,
This one works well.
I updated and pushed.
Thanks

PR #62, opened 2021-11-30
Mithun belief reader cleanup

mithunpaul08:


MihaiSurdeanu:
@mithunpaul08: what is this meant to do:

```
if (Try(word.toInt).isSuccess || Try(word.toFloat).isSuccess || Try(word.toDouble).isSuccess || Try(word.toLong).isSuccess ) {
         numberCounter = numberCounter + 1
       }
```

MihaiSurdeanu:
added @kwalcock

kwalcock:
It's trying to see if word is a number.  Things like toInt will throw an exception on an error, so Try.isSuccess seems like a good strategy.  Probably toLong and toDouble would suffice because those types are wider.  However, I wonder if the intended design was to look at POS or something else that might already have done the parsing.

mithunpaul08:
short version: the code in its current form will do two basic changes mihai had suggested 
- ignore sentences more than 150 tokens
- ignore sentences that have more than half of the tokens as numbers

however,
long version: yes, like @kwalcock said with that code i am just trying to find if a word is a number or not. And the only reason i added that code there is because @MihaiSurdeanu had added other filters in that page. But i also agree with keith that either the filtering should be done based on pos tags, or the filtering should happen even before the extraction occurs (why do mention extraction on a sentence if it is never going to be used). or mihai: comments? / did you want this to be done only in case of sentences which has beliefs.

MihaiSurdeanu:
Thanks both!
@mithunpaul08:
- This filter needs to be changed to use POS tags. 
- The filter needs to be moved in this method, which filters out sentences before extraction:
https://github.com/clulab/habitus/blob/main/src/main/scala/org/clulab/habitus/HabitusProcessor.scala#L30

kwalcock (commented):
It's getting better and better!  These filters are starting to look familiar, though.  There are some parts of Eidos that do the same sort of filtering, but they look at the text before processors has a chance to get to it because long and numbery sentences can take a long time to process.  I'm mentioning it to make sure that there wasn't a miscommunication somewhere.

mithunpaul08:
@MihaiSurdeanu While I was implementing the checkBadSentence using POS tokens at the [location](https://github.com/clulab/habitus/blob/main/src/main/scala/org/clulab/habitus/HabitusProcessor.scala#L30) in  habitusProcessor you suggested, I see that the sentence is not parsed/annotated by cluProcessor yet and hence doesn't have POS tags. The annotation doesn't happen until later when beliefProcessor.parse(). Is that a mistake in design choice or am i missing something?

kwalcock:
I don't know the answer, but in case it helps, here is where similar things are done in eidos: https://github.com/clulab/eidos/blob/master/eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/EidosProcessor.scala.  I remember it being a challenge to shoe horn it in there.

MihaiSurdeanu:
@mithunpaul08 : good point. To avoid this Catch 22 situation, I propose something simpler than @kwalcock's solution: mark any token that does not contain at least an alpha character as a potential number. Then proceeds with the heuristics we discussed, e.g., remove sentences that contain more 50% of these tokens (or some other threshold you observe). Thanks!

mithunpaul08:
FYI. These 3 test cases are getting ignored in org/clulab/beliefs/TestBeliefReader.scala
not sure it is by design. 

val sent13 = "There’s no reason to think that Mu is worse than Delta"
val sent18 = "Paul questions the popular theory that Trump is an idiot."
val sent19 = "Senator Harris mistrusts Trump’s medical recommendations but trusts Fauci’s"

PR #63, opened 2021-11-30
belief rule adjustments

maxaalexeeva:
- added belief rule and tests
- another isProposition check for discontinuous beliefs (might contain redundant code; needs to be simplified?)
- allowing believer to be optional in one of the rules 
- adding passing/failing test labeling


maxaalexeeva:
@MihaiSurdeanu this will most surely conflict with @mithunpaul08's branch. Let me know when you have merged that one and I can resolve.

src/test/scala/org/clulab/habitus/utils/Test.scala
```diff
@@ -3,4 +3,9 @@ package org.clulab.habitus.utils
 import org.scalatest.FlatSpec
 import org.scalatest.Matchers
 
-class Test extends FlatSpec with Matchers
+class Test extends FlatSpec with Matchers {
```
kwalcock, line 6: 👍

src/test/scala/org/clulab/habitus/utils/Test.scala
```diff
@@ -3,4 +3,9 @@ package org.clulab.habitus.utils
...
+  val passingTest = it
```
kwalcock, line 7: 👍

MihaiSurdeanu:
This is cool. Thanks @maxaalexeeva !

We should merge this soon because I want to make some changes to the belief package. @kwalcock, ok to merge?

src/main/scala/org/clulab/beliefs/BeliefProcessor.scala
```diff
@@ -78,6 +86,27 @@ class BeliefProcessor(val processor: Processor,
...
+    val sent = belief.sentenceObj
+    val tags = sent.tags.get
+    val beliefSpan = belief.tokenInterval
+    val beliefThemeSpan = beliefTheme.tokenInterval
+    var nounCount = 0
+    var verbCount = 0
+    for(i <- beliefSpan.start until beliefSpan.end) {
+      if(tags(i).startsWith("NN")) nounCount += 1
+      else if(tags(i).startsWith("VB")) verbCount += 1
+    }
```
kwalcock, line 101: Maybe

```scala
val nounCount = tags.view(beliefSpan.start, beliefSpan.end).count(_.startsWith("NN")) + ...
```

or a mini function that takes a tokenInterval and a prefix since there are 4 times that it happens.

```scala
def countTags(tokenInterval: TokenInterval, prefix: String): Int =
    tags.view(tokenInterval.start, tokenInterval.end).count(_.startsWith(prefix))

val nounCount = countTags(belief.tokenInterval, "NN") + countTags(beliefThemeSpan, "NN")
val verbCount = ...
```
Maybe it should take two intervals so that "NN" isn't repeated.

kwalcock:
I tried really hard and could only come up with one suggestion which was only cosmetic.  I haven't been the gatekeeper, not really knowing what all is going on.  I liked the categorization of tests.  Someone was paying extremely close attention.  I hope it doesn't turn out to be an antipattern of some kind (and if it does it will have been my fault).

MihaiSurdeanu:
so, merge?

kwalcock:
It's certainly OK with me.

kwalcock:
Implicature is a difficult.  I will go ahead and merge.

PR #64, opened 2021-12-02
Experimentally use ArrayView

kwalcock:


MihaiSurdeanu:
nice

kwalcock:
Please excuse the intrusion.  While working with Processors I noticed that Array.view, which is already used a couple of times in that project, is no longer available in Scala 2.13 or 3.  This is the same method that I had suggested using here.  For Processors I intend to replace Array.view with ArrayView , but it will be some time before it gets updated.  It seems useful to try it out here in the meantime.

PR #76, opened 2021-12-20
Address IntelliJ nags

kwalcock:


kwalcock:
I was double checking on UTF8 and file closing and decided to address IntilliJ's complaints.  Functionality should not have changed.  @mithunpaul08 is tagged so that I don't merge my own nonsense.

mithunpaul08:
ok done. noted.

PR #84, opened 2021-12-27
will print only jsonl files. no more json

mithunpaul08:
had spoken to Tomek and he said jsonl is enough for him. So I have removed the lines that print json. Though a part of me says more the better? I will let you make that call @MihaiSurdeanu

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ project/project/
 in/
 out/
 mentions.json
+mentions.jsonl
 mentions.tsv
 
 # docker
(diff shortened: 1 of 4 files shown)


kwalcock:
It looks fine (except for the failing test).  I suspect that an additional import can be removed.

MihaiSurdeanu:
I am Ok with JSONL only. But please comment out the failing JSON test.

kwalcock:
If "it" is replaced by "ignore", then the code will continue to compile and be refactored with other changes.  It sometimes works better than commenting things out.

kwalcock:
The tests pass, so I'll merge.

PR #85, opened 2021-12-27
Check for jsonl output

kwalcock:


kwalcock:
@mithunpaul08, maybe this will work.  It doesn't use the it/ignore thing, though.

MihaiSurdeanu:
Thanks @kwalcock ! Merged.

kwalcock:
It didn't fix the problem.  I'll have an update momentarily.

PR #86, opened 2021-12-27
Fix TestPrinting

kwalcock:


kwalcock:
The unit tests only run on the main branch, so that's why they weren't run last time on the server.

This test still isn't the best, but before it didn't erase the files it had created, so it could pass in the development environment and not pass on the server.  That's fixed now (I think) and one "it" was changed to "ignore".

mithunpaul08:
if this branch is an exact copy of mithun_remove_json, please merge this only @MihaiSurdeanu  and delete mithun_remove_json please

PR #90, opened 2022-01-05
belief extraction 

maxaalexeeva:
major changes:
- using a lexicon to find agents and only extract beliefs where the believer is an agent 
- adding a generic agent rule (personal pronouns) to allow beliefs with  non-specific agents 
- moving belief triggers to a separate file
- trying to first extract propositions and then attach them to belief triggers (supposed to help with expansion; to be investigated further)
- disabling a bunch of rules (produced too many non-beliefs)
- marking tests that are not exactly beliefs with `toDiscuss` (will be ignored)
- adding LikelyFact beliefs (`acknowledge`, `state`; disabled for now)

minor changes
- did some rule renaming
- added a few filter actions, e.g., don't allow mentions where belief and believer are on the same side of the trigger


maxaalexeeva:
Hi @MihaiSurdeanu, these are the changes in processors I made for habitus expansion: https://github.com/clulab/processors/commit/54a786f0a7dc5535a8a694ebffa47727dc9484fe
These changes do break some tests in processors (I have not investigated much yet), so we may want to have allowed expansion dependencies defined in habitus eventually. This habitus branch passes even without those changes/is mergeable, though, in case you think including those dependencies here can wait.

MihaiSurdeanu:
Thanks @maxaalexeeva! I'll take a look soon.

MihaiSurdeanu:
These are great! Thank you @maxaalexeeva !

MihaiSurdeanu:
@maxaalexeeva : Also, the changes in processors seem Ok to me... But can you please let me know which tests fail when you can? We should understand what happens before merging.

maxaalexeeva:
@MihaiSurdeanu 

Here are the failing tests:

```
[info] CustomizableRuleBasedEntityFinder
[info] - should expand using specified dependencies *** FAILED ***
[info]   Vector(org.clulab.odin.TextBoundMention@7172d7d3, org.clulab.odin.TextBoundMention@c467da5) had size 2 instead of expected size 3 (TestCustomizableRuleBasedEntityFinder.scala:19)
```
Then three tests in org.clulab.openie.entities.TestRuleBasedEntityFinder with the error message that includes this:

`java.lang.RuntimeException: sentence has no dependencies`

and this one:

```
[info] - should generate correct dependencies in test sentence 1 *** FAILED ***
[info]   false was not true (TestFastNLPProcessor.scala:22)
```

The last one makes most sense to me off the top of my head since I switched the set of dependencies from universal enhanced to hybrid (although, based on the documentation, hybrid includes universal enhanced, so I am not sure why the subset of deps inside the test would not be included in the hybrid as well).

I can look into what is going on with the tests tomorrow and/or Saturday.

MihaiSurdeanu:
Thanks @maxaalexeeva !
Not urgent, but when you have some time, can you please extract more details about these tests? You're right: the dependencies should not be empty, so this could be a bug.

PR #91, opened 2022-01-10
basic setup for interview reader + moving beliefs under habitus

maxaalexeeva:
- added an Interview reader with sample rules, where Allegra C. can add the rules for reading interviews
- moved belief reader to be under habitus
- moved lexicons from diff subproject to one dedicated directory


maxaalexeeva:
@kwalcock I am having trouble using Eidos here. I tried importing both SimpleEidos and creating a new EidosSystem, but I am having an issue with a Lum dependency:

```
Processing failed with the following error:
java.lang.NoClassDefFoundError: ai/lum/common/ConfigUtils$ConfigWrapper$
        at org.clulab.wm.eidos.components.ComponentsBuilder.<init>(ComponentsBuilder.scala:43)
        at org.clulab.wm.eidos.EidosSystem.<init>(EidosSystem.scala:43)
```
The ConfigWrapper is from Lum common 0.0.8, used by Eidos; habitus uses version 0.1.5 and for some reason really does not want to downgrade this dependency to 0.0.8 even if I redefine it in build. sbt. Any thoughts on how to fix this? 

Also, @kwalcock and @MihaiSurdeanu, I am having trouble with using MultiPrinter (don't know how to get it to write more than one type of mention), so I just use a standard PrintWriter.

kwalcock:
FWIW I'm looking at it.

maxaalexeeva:
@kwalcock, thank you!

kwalcock:
The problem with lum common can be bypassed, at least temporarily, but then things get messed up when loading odin, probably because the eidos that is doing the loading expects processors 8.4.5 rather than 8.4.7.  The rest of the project won't compile with the earlier version.  It may require either a general eidos update to the newer processors or a locally published eidos after all.  The lum common dependency might be updated at the same time.  It's going to have to wait until at least tomorrow, I think.

maxaalexeeva:
@kwalcock, thank you for investigating! Definitely can wait. It might make sense for me to meet with Allegra first to see more specifically what she needs to make sure we don't go into a major refactor needlessly?

kwalcock:
I'm updating eidos locally to see if it helps.

kwalcock:
It did help.  It will take a couple of hours for changes to propagate through the system.  They are very minor changes.

maxaalexeeva:
@kwalcock thank you!

kwalcock:
In about a minute there will be a PR on this branch with code that is working here in case it is useful.

maxaalexeeva:
@kwalcock thank you! Will look shortly!

maxaalexeeva:
@kwalcock, thank you! Works beautifully! I included the causation extractor in the interviews processor to make sure they make it both into the reader output and the shell. Hope that's alright.

@MihaiSurdeanu, I think this is everything I wanted to include for Allegra's initial setup (=pr is ready)

src/main/scala/org/clulab/habitus/interviews/InterviewsReader.scala
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  def example(): Unit = {
+    val text = "This is a test.  Water trucking in Ethiopia has decreased over August due to the cost of fuel."
+    val eidosSystem = SimpleEidos(useGeoNorm = false, useTimeNorm = false)
+    val annotatedDocument = eidosSystem.extractFromText(text)
+    val simpleSerializer = SimpleSerializer(annotatedDocument)
+
+    simpleSerializer.serialize()
+  }
```
kwalcock, line 29: This and the call to it can be removed now, probably with some imports as well.

maxaalexeeva: @kwalcock, good point, thanks! Done!

MihaiSurdeanu:
Looks good! Ok to merge?

maxaalexeeva:
I think so. @kwalcock?

PR #92, opened 2022-01-10
Show how eidos might be integrated

kwalcock:


kwalcock:
@maxaalexeeva, this runs for me up to where it tries to find files but can't with the "" value.

PR #93, opened 2022-01-26
Update sbt, scala

kwalcock:
Also experiment with ^C exit

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,5 +32,7 @@ libraryDependencies ++= {
 
 addCommandAlias("dockerize", ";compile;test;docker:publishLocal")
 
-// This seems to help with Ctrl+C processing.
-run / fork := true
+// This with true seems to help with Ctrl+C processing.
+// run / fork := true
+// However, it has the side-effect of causing the BeliefShell to fall through and exit. 
+run / fork := false
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 name := "habitus"
 organization := "org.clulab"
-scalaVersion := "2.12.10"
+scalaVersion := "2.12.15"
 
 lazy val core: Project = (project in file("."))
     .enablePlugins(JavaAppPackaging, DockerPlugin)
@@ -10,7 +10,7 @@ pomIncludeRepository := { (repo: MavenRepository) =>
 }
 
 // for processors-models
-resolvers += "Artifactory" at "http://artifactory.cs.arizona.edu:8081/artifactory/sbt-release"
+resolvers += ("Artifactory" at "http://artifactory.cs.arizona.edu:8081/artifactory/sbt-release").withAllowInsecureProtocol(true)
 // for ontologies related to eidos
 resolvers += "jitpack" at "https://jitpack.io"
 
@@ -31,3 +31,6 @@ libraryDependencies ++= {
 }
 
 addCommandAlias("dockerize", ";compile;test;docker:publishLocal")
+
+// This seems to help with Ctrl+C processing.
+run / fork := true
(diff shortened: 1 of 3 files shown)


kwalcock:
@maxaalexeeva, can you try within `sbt` the command `runMain org.clulab.habitus.ExitApp` and see if Ctrl+C works for you?  It should stop printing the numbers and return to the `sbt` prompt rather than exit `sbt`.

kwalcock:
I think that for a project being actively developed, we should be using the most current version of Scala that we can, so 2.12.15 for this.  I can't think of a good reason to use an earlier 2.12.X.  So, I've updated that.

`sbt` might be different, though.  The 1.2.6 that was there doesn't work correctly on my computer.  Eidos is at 1.4.5.  odinson is up to 1.6.0 and about to go 1.6.1, fatdynet was just incremented to 1.6.1, processors will be there shortly.  I've upped this to 1.6.1.  It will be slightly experimental, but this seems like a relatively good place for that.  `sbt` should update itself automatically, but there could be unforeseen consequences.

maxaalexeeva:
@kwalcock thanks for looking into this! This looks like the behavior you described?

```
[info] 45959
[info] 45960
[info] 45961
[info] 45962
[info] 45963
[info] 45964
[info] 45965
Cancelled: runMain org.clulab.habitus.ExitApp
[error] Cancelled: runMain org.clulab.habitus.ExitApp
[error] Use 'last' for the full log.
sbt:habitus> 
```
It does not seem like I can run the belief shell now, though (partial output):

```
[info] 22:56:02,198 |-INFO in ch.qos.logback.classic.joran.JoranConfigurator@4de5031f - Registering current configuration as safe fallback point
[error] [dynet] Loading DyNet from /tmp/libdynet_swig-407533715635910396.so...
[error] [dynet] random seed: 2522620396
[error] [dynet] allocating memory: 512MB
[error] [dynet] memory allocation done.
[info] Welcome to the Habitus Shell!
[info] COMMANDS:
[info]     :help   => show commands
[info]     :reload => reload the variable processor
[info]     :exit   => exit system
[info] (Habitus)>>> 
[info] Bye!
[success] Total time: 57 s, completed Jan 25, 2022 10:56:06 PM

```

kwalcock:
@maxaalexeeva, the first one looks OK.  They've changed the library they use for console I/O recently and perhaps it has messed up the shell.  I will try it.  Is this on the command line and Linux then?

maxaalexeeva:
Linux, running `./belief-shell` inside intellij Terminal

kwalcock:
Thanks.  Running inside IntelliJ is an extra level of difficulty.  It has never worked for Windows.  I will try Linux tomorrow.  It may be that the sbt version needs to be lowered.  I assume it was working correctly like this before?

On the other hand, it doesn't seem right to start up a GUI in order to run a CLI program inside a script, etc.  gui(bash(sbt(java(org.clulab.habitus.beliefs.BeliefShell)))).  Why not just right click on org.clulab.habitus.beliefs.BeliefShell and run it directly with a debugger and all?

maxaalexeeva:
Yeah, was working before. I can try running it other ways. I somehow just
ended up running everything through the terminal.

kwalcock:
@maxaalexeeva, it looks like that fork := true setting that helped with ^C messed up other input.  I guess that the forked process did not have access to stdin so that the program wouldn't wait for any input.  It is set to false now with some comments.

This worked on the command line and in the Terminal pane of IntelliJ, but a better place to run it is in the Console sub-pane of the Debug pane.  I just right click on the BeliefShell file and choose Debug and it will show up.   One needs to click in the Console pane to make it active before entering anything.  An advantage is that one can just click the stop button once or twice to end the program without any ^C.

kwalcock:
I'll probably merge this soon unless more problems are found.

maxaalexeeva:
@kwalcock I tried running the things that I need to run, and everything seems alright. Thanks!

PR #94, opened 2022-02-05
planting areas reader

maxaalexeeva:
Major changes:
- reader for planting areas
- variable reader accepts path to rule master file 
- separate context extractor (applied within processors, not readers, to allow printing of context attachments in the shell) 

Minor changes:
- disallowing negative area values (within reader)
- including percentages as values for planting areas
- splitting area mentions with multiple values into multiple mentions with a single value each
- applying global actions to variable extractions
- updates to context tests to account for the new context reader

### DIFF
--- a/src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
+++ b/src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
@@ -9,22 +9,26 @@ import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 
 trait Context extends Attachment {
-  def getArgs() = this.getClass.getDeclaredFields.toList
-    .map(i => {
-      i.setAccessible(true)
-      (i.getName, i.get(this))
+  def getArgValuePairs() = this.getClass.getDeclaredFields.toList
+    .map(arg => {
+      arg.setAccessible(true)
+      (arg.getName, arg.get(this))
     })
 
-  //get header as tsv string
-  // get context as tsv string
-  // seq[key, value] (because we want to keep preference on the order)
+  def getTSVContextHeader() = {
+    getArgValuePairs().map(_._1).mkString("\t")
+  }
 
+  def getTSVContextString() = {
+    getArgValuePairs().map(_._2).mkString("\t")
+  }
 }
 
 case class DefaultContext(location: String, date: String, crop: String, fertilizer: String, comparative: Int) extends Context
 
-
 trait ContextExtractor {
+
+  def getContextPerMention(mentions: Seq[Mention], entityHistogram: Seq[EntityDistFreq], doc: Document, label: String): Seq[Mention]
 
   def getComparative(mention: Mention): Int = {
     val relative = Seq("vs", "vs.", "respectively")
@@ -34,7 +38,9 @@ trait ContextExtractor {
   def getLocation(m: Mention, thisSentLocs: Seq[Mention], frequencyContext: Map[Int, ContextDetails]): String = {
     val location = thisSentLocs.length match {
       // if no locations in sentence, use the most freq one in +/- 1 sent window
-      case 0 => frequencyContext(m.sentence).mostFreqLoc1Sent
+      case 0 => if (frequencyContext.nonEmpty) {
+        frequencyContext(m.sentence).mostFreqLoc1Sent
+      } else "N/A"
       case 1 => thisSentLocs.head.text
       case _ => {
         val nextLoc = findClosestNextLocation(m, thisSentLocs)
@@ -45,17 +51,23 @@ trait ContextExtractor {
   }
 
   def getCropContext(m: Mention, frequencyContext: Map[Int, ContextDetails]): String = {
-    frequencyContext(m.sentence).mostFreqCrop0Sent
+    if (frequencyContext.nonEmpty) {
+      frequencyContext(m.sentence).mostFreqCrop0Sent
+    } else "N/A"
   }
 
   def getFertilizerContext(m: Mention, frequencyContext: Map[Int, ContextDetails]): String = {
-    frequencyContext(m.sentence).mostFreqFertilizer1Sent
+    if (frequencyContext.nonEmpty) {
+      frequencyContext(m.sentence).mostFreqFertilizer1Sent
+    } else "N/A"
   }
 
   def getDate(m: Mention, thisSentDates: Seq[Mention], frequencyContext: Map[Int, ContextDetails]): String = {
     // if no dates in sentence, use the most freq one in +/- 1 sent window
     val date = thisSentDates.length match {
-      case 0 => frequencyContext(m.sentence).mostFreqDate1Sent // because there are no dates in this sentence, take most freq in -/+ 1 window
+      case 0 => if (frequencyContext.nonEmpty) {
+        frequencyContext(m.sentence).mostFreqDate1Sent
+      } else "N/A"// because there are no dates in this sentence, take most freq in -/+ 1 window
       case 1 => thisSentDates.head.text
       case _ => findClosestDate(m, thisSentDates).text
     }
@@ -63,7 +75,6 @@ trait ContextExtractor {
   }
 
   def compressContext(doc: Document, allEventMentions: Seq[Mention], entityHistogram: Seq[EntityDistFreq]): mutable.Map[Int, ContextDetails] = {
-    //    println("COMPRESSING CONTEXT")
     //sentidContext is a data structure created just to carry contextdetails to the code which writes output to disk
     //note: value=Seq[contextDetails] because there can be more than one mentions in same sentence
     val sentidContext = mutable.Map[Int, ContextDetails]()
@@ -95,7 +106,6 @@ trait ContextExtractor {
   def findMostFreqContextEntitiesForAllEvents(mentionContextMap: mutable.Map[Mention, Seq[EntityDistFreq]], howManySentAway:Int, entityType:String):Seq[MostFreqEntity] = {
     mentionContextMap.keys.toSeq.map(key=>findMostFreqContextEntitiesForOneEvent(key,mentionContextMap(key), entityType,howManySentAway))
   }
-
 
   //if key exists add+1 to its value, else add 1 as its value
   def checkAddFreq(map: mutable.Map[String, Int], key: String, freq: Int): Int = {
--- a/src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
+++ b/src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
@@ -1,12 +1,14 @@
 package org.clulab.habitus.utils
 
 import org.clulab.habitus.variables.EntityDistFreq
+
 import org.clulab.odin.{Attachment, Mention}
 import org.clulab.processors.Document
 
+import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 
-case class Context(location: String, date: String) extends Attachment {
+trait Context extends Attachment {
   def getArgs() = this.getClass.getDeclaredFields.toList
     .map(i => {
       i.setAccessible(true)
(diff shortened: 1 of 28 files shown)


maxaalexeeva:
Hey @MihaiSurdeanu and @kwalcock! This is me trying to refactor variable reading to have things more modular. 

Note: this is very much a draft, so please excuse wrong paths and println's. 

There are a lot of changes and I am not sure if looking at the diff is very helpful, so here is the key point:  There is a separate ContextExtractor. Before my changes, the context was determined and output within the printers. I think it is easier if context is extracted by context extractor and attached to each mention as an Attachment---that makes printers more reusable in case different variable readers need different context (e.g., only planting areas include info on whether or not the values are being compared to something else and (for now) only planting dates include info on crops and fertilizers).

If this is not great, we should probably talk more and decide more specifically how you want things organized and we can discard this. *I also broke some tests, but have not looked at them yet because I don't know if this refactoring is gonna stay.

MihaiSurdeanu:
Hi @maxaalexeeva,
I think this is massive step forward! A few suggestions:
- The ContextExtractor should be a trait, and we should inherit from that to define specific context extractors for each readers
- We should not have absolute paths committed in the code. I think the InterviewsReader has some. They should be passed in the cmd line. We have this class to parse cmd line parameters: https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/utils/StringUtils.scala#L21 
- I think the old context code (mostFreq in same sentence, 1 sentence apart, etc.) can be removed. In fact, I think the two readers can (and should!) share the same ContextExtractor that is based mostly on yours.

maxaalexeeva:
Thanks @MihaiSurdeanu! Questions:


>     The ContextExtractor should be a trait, and we should inherit from that to define specific context extractors for each readers

I think I could use a bit more info on the structure. So, we can have a trait ContextExtractor, which can contain reusable methods that can be used as needed in more specific context extractors? 

>     I think the old context code (mostFreq in same sentence, 1 sentence apart, etc.) can be removed. 

Remove frequency-based context even for cases when there is no context within the sentence? Or just remove the same sentence portion for date/time? Also, frequency-based context is used for crops and fertilizers---we want to keep that, right? 

> In fact, I think the two readers can (and should!) share the same ContextExtractor that is based mostly on yours.
Allegra was mentioning that knowing more context about planting areas (e.g., what crop is involved) is useful, so planting reader should have at least all the context that was available for dates, right? 

Two more questions:
- are you okay with context stored in attachments to mentions?
- if I want to discuss the code a bit more, should I schedule a meeting or should we discuss this in the general lab/project meeting on Tuesday?

maxaalexeeva:
Hi @MihaiSurdeanu, I think this is mainly ready for review, but it depends on the current snapshot of processors (the percentages). How to proceed?

MihaiSurdeanu:
Thanks @maxaalexeeva ! I'll take a look soon.

MihaiSurdeanu:
@maxaalexeeva: this is overall great. A few comments:

- Update processors to 8.4.8 and push, to check that tests are passing.

- In `getDate`, if there are no dates in +/- 1 sentences, I think you should continue to backup to +/- 2, 3, etc., until let's say +/- 10 sentences apart.

- Hard to see from the diff, but I think we should not be using `compressContext`. This method produces too much information. We decided to switch to your simpler (and, hopefully, more informative) context, right? If we're not using it, I propose to remove all that code (including its auxiliary methods) so we keep our code base clean.

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
```diff
@@ -0,0 +1,260 @@
...
+
+case class DefaultContext(location: String, date: String, crop: String, fertilizer: String, comparative: Int) extends Context
+
+trait ContextExtractor {
+
+  def getContextPerMention(mentions: Seq[Mention], entityHistogram: Seq[EntityDistFreq], doc: Document, label: String): Seq[Mention]
+
+  def getComparative(mention: Mention): Int = {
+    val relative = Seq("vs", "vs.", "respectively")
+    if (mention.sentenceObj.words.intersect(relative).nonEmpty) 1 else 0
```
kwalcock, line 35: If this runs lots of times and efficiency is important, instead of calculating the exact intersection, one might look for a single match and stop with
```scala
relative.exists(mention.sentenceObj.words.contains(_))
```
It's not as pretty, granted.  I guess that the Seq could be a simple Array and it could be stored statically in a companion object somewhere, in case it is being recreated for each call.

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
```diff
@@ -0,0 +1,260 @@
...
+      (arg.getName, arg.get(this))
+    })
+
+  def getTSVContextHeader() = {
+    getArgValuePairs().map(_._1).mkString("\t")
+  }
+
+  def getTSVContextString() = {
+    getArgValuePairs().map(_._2).mkString("\t")
+  }
```
kwalcock, line 24: It *seems* like the argValuePairs are generic enough.  Someone else will want CSV and someone else will somehow slip tabs or commas into the data so that the output is messed up.  It could be better to handle that part in the specific printers.

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
```diff
@@ -0,0 +1,260 @@
...
+  def getLocation(m: Mention, thisSentLocs: Seq[Mention], frequencyContext: Map[Int, ContextDetails]): String = {
+    val location = thisSentLocs.length match {
+      // if no locations in sentence, use the most freq one in +/- 1 sent window
+      case 0 => if (frequencyContext.nonEmpty) {
+        frequencyContext(m.sentence).mostFreqLoc1Sent
+      } else "N/A"
+      case 1 => thisSentLocs.head.text
+      case _ => {
+        val nextLoc = findClosestNextLocation(m, thisSentLocs)
+        if (nextLoc != null) nextLoc.text else "N/A"
```
kwalcock, line 47: The null siren just went off.  Can this be made an Option instead?

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
```diff
@@ -0,0 +1,260 @@
...
+      frequencyContext(m.sentence).mostFreqFertilizerOverall
+    } else "N/A"
+  }
+
+  def getDate(m: Mention, thisSentDates: Seq[Mention], frequencyContext: Map[Int, ContextDetails]): String = {
+    // if no dates in sentence, use the most freq one in +/- 1 sent window
+    val date = thisSentDates.length match {
+      case 0 => if (frequencyContext.nonEmpty) {
+        frequencyContext(m.sentence).mostFreqDate1Sent
+      } else "N/A"// because there are no dates in this sentence, take most freq in -/+ 1 window
```
kwalcock, line 70: I'd be likely to mistype one of these or decide to change it one day and miss one, therefore I'm forced to write
```scala
val NA = "N/A"
```

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
```diff
@@ -0,0 +1,260 @@
...
+    //go through each of the contexts and find if any of them satisfies the condition in the query
+    for (ctxt <- contexts) {
+      for (sentFreq <- ctxt.entityDistFrequencies) {
+        val sentDist = sentFreq._1
+        val freq = sentFreq._2
+        if (sentDist <= howManySentAway && ctxt.nerTag.contains(entityType)) {
+          val newFreq = checkAddFreq(entityFreq, ctxt.entityValue, freq)
+          if (newFreq >= maxFreq) {
+            maxFreq = newFreq
+            mostFreqEntity = ctxt.entityValue
```
kwalcock, line 133: An alternative, not necessarily better, is
```scala
val mostFrequentEntity = if (entityFreq.nonEmpty) Some(entityFreq.maxBy(_._2)._1) else None
```
There are fewer special values like 0 and "".  Later Scalas have maxByOption to take care of the empty condition.

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
```diff
@@ -0,0 +1,260 @@
...
+    }
+  }
+  //if none, return "N/A", else return String value of entity e.g.:"SENEGAL"
+  def checkIfEmpty(mostFreqEntity: Option[String]): String = mostFreqEntity.getOrElse("N/A")
+
+  def findClosestNextLocation(mention: Mention, locations: Seq[Mention]): Mention = {
+    if (locations.length == 1) return locations.head
+    val nextLocations = locations.filter(_.tokenInterval.start > mention.arguments("value").head.tokenInterval.start)
+    if (nextLocations.nonEmpty) nextLocations.minBy(_.tokenInterval)
+    else null
```
kwalcock, line 246: So here I'd very much recommend returning Option[Mention] as standard practice, for the benefit of others.

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  def findClosestDate(mention: Mention, dates: Seq[Mention]): Mention = {
+    var minDist = 100
+    var minDistDate = dates.head
+    for (date <- dates) {
+      if (math.abs(mention.tokenInterval.start - date.tokenInterval.end ) < minDist | math.abs(mention.tokenInterval.end - date.tokenInterval.start) < minDist) minDistDate = date
```
kwalcock, line 253: || ?

kwalcock: Would dates.minBy work here?  That 100 might lead to an incorrect answer, I think.

kwalcock (commented):
There's lots and lots of really good code here, so good that I started to get nit-picky.  Many comments can be ignored.  I'm unaware of any big picture problems but haven't finished, either.  It is interesting reading.  I learned a cool trick!

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
```diff
@@ -0,0 +1,260 @@
...
+
+import scala.collection.mutable
+import scala.collection.mutable.ArrayBuffer
+
+trait Context extends Attachment {
+  def getArgValuePairs() = this.getClass.getDeclaredFields.toList
+    .map(arg => {
+      arg.setAccessible(true)
+      (arg.getName, arg.get(this))
+    })
```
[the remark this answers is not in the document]
maxaalexeeva: @kwalcock thanks for all the comments! This I, admittedly, found somewhere on stack overflow. Seems to be doing what I want!

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
maxaalexeeva: Yes!

src/main/scala/org/clulab/habitus/utils/JsonPrinter.scala
```diff
@@ -40,47 +39,21 @@ class JsonPrinter(outputFilename: String) extends Printer {
...
+      ("variableText" -> variableText) ~
+        ("valueText" -> valueText) ~
+        ("valueNorm" -> valueNorm) ~
+        ("sentenceText" -> sentenceText) ~
+        ("inputFilename" -> inputFilename)
+
+    // add key-value pairs from the context attachment
+    mention.attachments.head.asInstanceOf[Context].getArgValuePairs()
+      .foreach {
+        i =>
```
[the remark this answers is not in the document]
kwalcock: I've been thinking about this.  Isn't a side-effect of `getArgValuePairs` and then `toString` that the results are all strings, even if they are numbers like context.comparative might be?  If so, it might fool a consumer of the output.

I'm not happy with my code below, but something like it might be necessary:
```scala
    // This should be reusable across JPrinters.
    def toJObject(argValuePairs: List[(String, AnyRef)]): JObject = {
      new JObject(argValuePairs.map { case (arg, value) => JField(arg,
        value match {
          case l: java.lang.Long => JLong(l)
          case i: java.lang.Integer => JInt(BigInt(i))
          case s: String => JString(s)
        }
      )})
    }

    val jObject1: JObject =
      ("variableText" -> variableText) ~
        ("valueText" -> valueText) ~
        ("valueNorm" -> valueNorm) ~
        ("sentenceText" -> sentenceText) ~
        ("inputFilename" -> inputFilename)
    val jObject2: JObject = toJObject(mention.attachments.head.asInstanceOf[Context].getArgValuePairs())
    val jObject = jObject1 ++ jObject2
```

kwalcock (commented):
Tiny crumbs of food for thought...

maxaalexeeva:
@kwalcock thanks for reviewing! I am now going through the big-scale changes since I already started on those and will then address your feedback!

src/test/scala/org/clulab/habitus/variables/TestContextExtractor.scala
```diff
@@ -23,9 +22,9 @@ In Senegal maturity  came in early November ( Tab.I ) for 1995.
 In Senegal maturity  came in early November ( Tab.I ) for 1995.
 """
 
-  def getMSFreq(text: String): Seq[vr.MostFreqEntity] = {
+  def getMSFreq(text: String): Seq[ce.MostFreqEntity] = {
```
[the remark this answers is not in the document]
maxaalexeeva: Based on @MihaiSurdeanu 's comment from yesterday, I think MostFreqEntity might go away altogether.

src/main/scala/org/clulab/habitus/utils/JsonPrinter.scala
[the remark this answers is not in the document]
maxaalexeeva: > If so, it might fool a consumer of the output.

Great point!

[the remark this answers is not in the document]
maxaalexeeva: Hm, looks like adding objects like that results in a different json structure, e.g.,

```
[ {
    "variableText" : "area",
    "valueText" : "35%",
    "valueNorm" : "35 %",
    "sentenceText" : "Financing of agricultural production  Credit committees ( 1er and 2th ) of La Banque Agricole Zone Nord met and , the volume of credit retained for the Financing of  Agricultural Production for the 2021_2022 rainy season , amounts to 3.635 billion CFA francs corresponding to an area equal to 11 268  hectares , i.e. a decrease of 35 % compared to the overall amount of credit granted to finance the production of the  wintering campaign 2020/2021 .",
    "inputFilename" : "Bulletin_24_08_2021.fr.en.txt"
  }, {
    "location" : "2021_2022",
    "date" : "north zone",
    "process" : "credit",
    "crop" : "yam",
    "fertilizer" : "N/A",
    "comparative" : 0
  } ]
```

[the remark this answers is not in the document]
maxaalexeeva: Never mind, after using toJObject, I do this and it works fine:
```
for (value <- argJObject.obj) {
      jObject = jObject ~ (value._1 -> value._2)
    }
```

maxaalexeeva:
Hey @MihaiSurdeanu and @kwalcock, I addressed your feedback and added some things Allegra requested. I expect github should give you the option to "review changes since last review"?

I switched to finding context by first finding relevant entities as mentions. that way i dont have to use entity histogram and compress context. I tried to reuse the histogram code, but it was getting too convoluted.

getProcess (Allegra wants to know what the text is vaguely about) is not great, but is a start. If each sentence only had one of possible processes (e.g., harvest or planting), I would say we could do vector similarity, but since that's not true, I might want to try to find that info same way as i do with other entities---extract relevant words as mentions and find closest to the area-value mentions. Open to suggestions and also that should probably be a different pull request, otherwise it will get delayed more.

The context tests are now disabled because they are mainly irrelevant now. If the way I extract context now is fine, then I will redo the tests to account for the new way handle context.

kwalcock (approved):

PR #97, opened 2022-03-07
Masha march7

maxaalexeeva:
- fixing area rules + action to filter out mentions where var and val are too far away (those are likely false positives)
- disabling belief rules that capture mainly author beliefs until we decide what to do with those (they might be useful for interview reading, but not for scientific publications/reports reading) 
- more work on extracting propositions to use as beliefs
- adjusting the tests to take these changes into account

### DIFF
--- a/src/test/scala/org/clulab/habitus/beliefs/TestBeliefReader.scala
+++ b/src/test/scala/org/clulab/habitus/beliefs/TestBeliefReader.scala
@@ -103,8 +103,9 @@ class TestBeliefReader extends Test {
     m.arguments("belief").head.text should be ("Tröegs might know a thing or two about beer")
   }
 
+  // Tests sent10 - sent_10_10 are author beliefs, which we are not extracting now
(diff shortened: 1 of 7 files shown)


MihaiSurdeanu:
Thanks! Merged.

src/test/scala/org/clulab/habitus/beliefs/TestBeliefReader.scala
```diff
@@ -164,7 +165,7 @@ class TestBeliefReader extends Test {
   }
 
   val sent10_5 = "It's plausible that Rudy Giuliani broke the law"
-  sent10_5 should "contain one belief" in {
+  ignore should s"contain one beliefin ${sent10_5}" in {
```
kwalcock, line 168: It's trivial, but there are several "beliefin"s.

kwalcock (approved):
The PR might have been merged already, so a new one might be necessary.

maxaalexeeva:
@kwalcock thank you! opened a new pr!

PR #99, opened 2022-03-21
Habitus Reader

maxaalexeeva:
One reader to rule them all

### DIFF
--- a/src/main/scala/org/clulab/habitus/utils/JsonPrinter.scala
+++ b/src/main/scala/org/clulab/habitus/utils/JsonPrinter.scala
@@ -48,11 +48,12 @@ class JsonPrinter(outputFilename: String) extends Printer {
         ("sentenceText" -> sentenceText) ~
         ("inputFilename" -> inputFilename)
 
-    // make an obj from key-value pairs from the context attachment
-    val argJObject: JObject = toJObject(mention.attachments.head.asInstanceOf[Context].getArgValuePairs())
-
-    for (value <- argJObject.obj) {
-      jObject = jObject ~ (value._1 -> value._2)
+    val context = mention.attachments.headOption
+    val argJObject: Option[JObject] = if (context.isDefined) Some(toJObject(context.get.asInstanceOf[Context].getArgValuePairs())) else None
+    if (argJObject.isDefined) {
+      for (value <- argJObject.get.obj) {
+        jObject = jObject ~ (value._1 -> value._2)
+      }
     }
 
     val json = stringify(jObject, pretty = true)
--- a/src/main/scala/org/clulab/habitus/utils/TsvPrinter.scala
+++ b/src/main/scala/org/clulab/habitus/utils/TsvPrinter.scala
@@ -34,11 +34,12 @@ class TsvPrinter(outputFilename: String) extends Printer {
         // ``variable`` and ``value`` we access the two through ``arguments`` attribute of the Mention class.
         m =>
            {
-            val varText = m.arguments(printVars.mentionType).head.text
-            val value = m.arguments(printVars.mentionExtractor).head
-            val valText = value.text
+            val variable = m.arguments(printVars.mentionType).headOption
+            val varText = if (variable.isDefined) variable.head.text else na
+            val value = m.arguments(printVars.mentionExtractor).headOption
+            val valText = if (value.isDefined) value.head.text else na
             val sentText = m.sentenceObj.getSentenceText
-            val valNorms = value.norms
+            val valNorms = if (value.isDefined) value.get.norms else None
             val norm = {
               if (valNorms.isDefined && valNorms.get.size >= 2) {
                 valNorms.filter(_.length >= 2).get(0)
@@ -48,10 +49,10 @@ class TsvPrinter(outputFilename: String) extends Printer {
                 //   For example, DATEs have norms, but CROPs do not
                 // in the latter case, we revert to the lemmas or to the actual text as a backoff
                 //
-                if (value.words.nonEmpty) {
-                  value.words.mkString(" ")
+                if (value.get.words.nonEmpty) {
+                  value.get.words.mkString(" ")
                 } else {
-                  value.text
+                  value.get.text
                 }
               }
             }
--- a/src/main/scala/org/clulab/habitus/utils/TsvPrinter.scala
+++ b/src/main/scala/org/clulab/habitus/utils/TsvPrinter.scala
@@ -56,8 +56,11 @@ class TsvPrinter(outputFilename: String) extends Printer {
                 }
               }
             }
-              val contextString = m.attachments.head.asInstanceOf[Context].getTSVContextString
-            pw.println(s"$varText\t$valText\t$norm\t$sentText\t$filename\t$contextString")
+             // this is for cases when there is no norm
+            val normString = if (norm.nonEmpty) norm + "\t" else ""
+            val context = m.attachments.headOption
+            val contextString = if (context.isDefined) context.get.asInstanceOf[Context].getTSVContextString else ""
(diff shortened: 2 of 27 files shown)


maxaalexeeva:
Hey @MihaiSurdeanu and @kwalcock,

There are a lot of files changed here (and this is very much work in progress), but based on the three files listed below as an example of how habitus reader would work, is this more or less what we are looking for for the default reader?

src/main/scala/org/clulab/habitus/variables/PlantingAreaReader.scala 
src/main/resources/application.conf
src/main/scala/org/clulab/habitus/HabitusReader.scala

Thanks!

kwalcock:
It looks well-organized.  I'll study it carefully.

maxaalexeeva:
Thanks, @kwalcock! I have not done a clean-up pass and also haven't adjusted all the other readers to use the new habitus one. I will adjust the other readers if you approve the way I did the planting area one. 

Also, I am including factuality config for future, but factuality is not included yet (@mithunpaul08 is working on that).

maxaalexeeva:
One other thing: current multi printer takes PrintVariables which defines the args a mention should have. In belief mentions, however, arguments can be different (e.g., sometimes there is a believer and sometimes not). Do you have any ideas on how to adjust the multi printer such that it can print a mention with any types of arguments? (I have some thoughts but none of them seem elegant)

kwalcock:
That's interesting inheritance: object PlantingAreaReader -> class HabitusReader -> trait App.  If it works out, I may need to steal it, because it looks really good.

The PlantingAreaReader.args array along with mentionArgs.head and last, kind of loses some useful documentation.  I wonder if we should just go all out literal and pedantic and in the config file write
```
    printVariables {
        mentionLabel = "Assignment"
        mentionType = "variable"
        mentionExtractor = "value"
    }
```
I *think* there is a way to read an object (like PrintVariables) directly from a config file.  We never use it, but maybe we should, eventually.  The individual variables can still be picked out for now, unless someone is curious.  (Now I'm curious and will look it up.)

I was trying to figure out the return type of this, especially given the two Seqs that look exactly the same:
```scala
  def parse(text: String): (Document, Seq[Mention], Seq[Mention])
```
The code that access the Tuple3 is opaque with _._1 and things, so I think this would be a good place for a case class, even if it is called a nondescript ParseResult, it could still have named fields of e.g. document, somethingMentions, somethingElseMentions, at least if those last two are consistent between processors.

More shortly...

kwalcock:
For the `PrintVariables`, is there a standard procedure for what to do if code like this fails either on the `apply()` or on `head`:
```scala
    val variableText = mention.arguments(printVars.mentionType).head.text
    val valueMention = mention.arguments(printVars.mentionExtractor).head
```
If those lines should always print "N/A" or should always be suppressed, then the printer might handle them.  There is, for example, a filter
```scala
        .filter { mention => mention.label.matches(printVars.mentionLabel) }
```
that could check if the value is present.  Otherwise there could be an `arguments.getOrElse()` and/or `headOption.getOrElse("N/A")" or something.  If there is extra complicated logic, then the caller might better filter out those mentions before handing off the job to the printer.

I should someday move those printers and PrintVariables to their own namespace.

kwalcock:
Not that I know what these beliefs, interviews, and variables are about, but those were my observations.  Some little complications are almost certain to appear as the others are reorganized.

maxaalexeeva:
> I was trying to figure out the return type of this, especially given the two Seqs that look exactly the same:
> 
>   def parse(text: String): (Document, Seq[Mention], Seq[Mention])
> 
> The code that access the Tuple3 is opaque with _._1 and things, so I think this would be a good place for a case class, even if it is called a nondescript ParseResult, it could still have named fields of e.g. document, somethingMentions, somethingElseMentions, at least if those last two are consistent between processors.

I am not 100% sure right now if it is critical we return two groups of mentions. It might be mainly for a more informative printout in the shell. Some places I just added them for consistency with the generic processor. I will investigate more.

kwalcock:
It is possible to do something like this
```scala
  val localConfig: Config = config[Config]("PlantingAreaReader")
  val masterResource: String = localConfig[String]("masterResource")
  val printVariables = ConfigBeanFactory.create(localConfig.getConfig("printVariables"), classOf[PrintVariables])
```
given
```scala
  case class PrintVariables(@BeanProperty var mentionLabel: String, @BeanProperty var mentionType: String, @BeanProperty var mentionExtractor: String) {
    def this() = this("", "", "")
  }

```
I'm not sure it's worth it, but here it is for the record.

maxaalexeeva:
@kwalcock 

> For the PrintVariables, is there a standard procedure for what to do if code like this fails either on the arguments() or on head:
> 
>     val variableText = mention.arguments(printVars.mentionType).head.text
>     val valueMention = mention.arguments(printVars.mentionExtractor).head 

I added a check in one of the printers, but I think it should be some methods shared between all of them. If you don't mind, I'd rather work on that in a separate PR (same goes for making a separate method for getting norms).

Also, @kwalcock and @MihaiSurdeanu: the interview reader is kept as is for now because there are many possible label mentions extracted there, so I will need to come up with some way to pass multiple possible PrintVariables to printers. (Also, as a separate pr if that's ok)

src/main/scala/org/clulab/habitus/utils/JsonPrinter.scala
```diff
@@ -48,11 +48,12 @@ class JsonPrinter(outputFilename: String) extends Printer {
         ("sentenceText" -> sentenceText) ~
         ("inputFilename" -> inputFilename)
 
-    // make an obj from key-value pairs from the context attachment
-    val argJObject: JObject = toJObject(mention.attachments.head.asInstanceOf[Context].getArgValuePairs())
-
-    for (value <- argJObject.obj) {
-      jObject = jObject ~ (value._1 -> value._2)
+    val context = mention.attachments.headOption
```
kwalcock, line 51: In a different iteration, I may include a method to perform this conversion and then add it to some shared superclass of JsonPrinter and JsonlPrinter along with the toJObject and anything else they share.

src/main/scala/org/clulab/habitus/utils/TsvPrinter.scala
```diff
@@ -34,11 +34,12 @@ class TsvPrinter(outputFilename: String) extends Printer {
         // ``variable`` and ``value`` we access the two through ``arguments`` attribute of the Mention class.
         m =>
            {
-            val varText = m.arguments(printVars.mentionType).head.text
-            val value = m.arguments(printVars.mentionExtractor).head
-            val valText = value.text
+            val variable = m.arguments(printVars.mentionType).headOption
```
kwalcock, line 37: Perhaps in a different iteration I would add an Opt or maybe to this like variableOpt or maybeVariable.  Not everyone agrees: https://stackoverflow.com/questions/11887181/scala-naming-convention-for-options.

kwalcock (approved):
It looks really, really good.  I wouldn't change anything in this PR but maybe keep some things in mind for the next one.

maxaalexeeva:
@kwalcock thanks for the feedback! Do I merge?

src/main/scala/org/clulab/habitus/utils/TsvPrinter.scala
kwalcock: I'd go for Opt to match other CLU Lab code, though.  There isn't a standard Maybe class in Scala (yet).  Sometimes one goes along with true and false.

kwalcock:
Sure, if you don't think Mihai will veto something.  It can be undone. if necessary.

MihaiSurdeanu:
please go ahead!

maxaalexeeva:
Hm... I guess I can wait until @MihaiSurdeanu has approved?

PR #100, opened 2022-03-29
Account for Option when printing

kwalcock:


kwalcock:
@mithunpaul08, it got a little complicated, but this is how I would do it.  I don't know what side effects the code might have, so please look carefully at the changes.

Also look for this comment and remove the factuality stuff.  Perhaps some change is missing from the build.sbt file.

// TODO: These were added for testing.  Remove afterwards!

PR #101, opened 2022-03-29
Use sentinel values for None

kwalcock:


kwalcock:
@mithunpaul08, I *think* this is the best place to change the verb etc. to N/A.

kwalcock:
I pretty sure this is no longer applicable.  I have a copy if that should change.

PR #103, opened 2022-03-30
Organize printer code

kwalcock:


kwalcock:
This might break some things because the logic for the TSV and Json printers was slightly different and shouldn't have been.  Some things were taken from the one and some from the other so that there is just one version.

maxaalexeeva:
@kwalcock, thank you! I have tried this out. Comments:

1) I think we need to adjust norm. It looks like this first check does not eliminate empty norms (I guess they are defined, but empty?), so in the belief reader tsv (where the "value" is not numeric), the third column is just empty:

https://github.com/clulab/habitus/blob/548c3709045158e1c1732d3a713e0342770ca6ab/src/main/scala/org/clulab/habitus/printer/ArgumentInfo.scala#L15 

If I add this `&& valueNormsOpt.get.head.nonEmpty`, then the else if kicks in, and the norm becomes the text of the argument, which means there is the same text the second and third column. Maybe we can have norm only show up if it contains some numbers and in other cases have "N/A"? Are there any non-numeric norms that could be useful to us? 

2) This does not address the issue of different kinds of mentions extracted by one reader (that is mentions with different labels). I have an idea on what to do with that, but maybe we can merge this the way it is now and I will get around to it some time soon in a different pr?

kwalcock:
Thanks, @maxaalexeeva.  I have little idea what is in these strings.  I'm not sure why there needs to be >= 2 norms, either.  You found the precise place where the issue is decided, though, and it should be easy to change.

kwalcock:
At some point I'll put all the N/A's in one place.

kwalcock:
Someone be sure to let me know when this breaks something.  I'm aware that this code will conflict with changes that people have in branches.  I can deal with the merge conflicts.

PR #104, opened 2022-04-07
Print all the arguments

kwalcock:


kwalcock:
JSON output looks like this:
```json
[
  {
    "sentenceText" : "Farmers ’ sowing dates ranged from 14 to 31 July for the WS and from 3 to 11 March for the DS .",
    "inputFilename" : "filename",
    "context" : {
      "location" : "N/A",
      "date" : "N/A",
      "process" : "planting",
      "crop" : "N/A",
      "fertilizer" : "N/A",
      "comparative" : 0
    },
    "arguments" : [ {
      "name" : "value",
      "text" : "from 14 to 31 July",
      "norm" : "XXXX-07-14 -- XXXX-07-31"
    }, {
      "name" : "variable",
      "text" : "sowing dates",
      "norm" : "N/A"
    } ]
  },
  {
```

kwalcock:
TSV looks like this:
```
sentenceText	inputFilename	location	date	process	crop	fertilizer	comparative	arg0_name	arg0_text	arg0_norm	...
Farmers ’ sowing dates ranged from 14 to 31 July for the WS and from 3 to 11 March for the DS .	filename	N/A	N/A	planting	N/A	N/A	0	value	from 14 to 31 July	XXXX-07-14 -- XXXX-07-31	variable	sowing dates	N/A
```
Because the printer doesn't see all Mentions before needing to print, the headers at the top are only approximate.

maxaalexeeva:
@kwalcock sorry about the delay. reading now.

maxaalexeeva:
@kwalcock 

> Because the printer doesn't see all Mentions before needing to print, the headers at the top are only approximate.

Just for the tsv, what if we just print each mention type (or more like mentions with the same set of arguments) to a different tsv?

kwalcock:
@maxaalexeeva, that's a really clever idea.  So if all the arguments turned out to be value and variable, one might get an output flle named output-value_variable.tsv with content

```
sentenceText	inputFilename	location	date	process	crop	fertilizer	comparative	value_text	value_norm	variable_text	variable_norm
Farmers ’ sowing dates ranged from 14 to 31 July for the WS and from 3 to 11 March for the DS .	filename	N/A	N/A	planting	N/A	N/A	0	from 14 to 31 July	XXXX-07-14 -- XXXX-07-31	sowing dates	N/A
```
If the arguments were inconsistent, there would be multiple output files with different argument names and different columns.  Would that be a problem?

I'd probably call this something like a DynamicTsvPrinter and add it as one of the choices.

You all are the customer and it sounds fine to me.  Can anyone second the idea, potentially @MihaiSurdeanu.

maxaalexeeva:
Adding a use case for this: Belief mentions can have two kinds of arguments:

believer and belief, e.g., farmers believe that chemicals are dangerous. 
beliefTheme and belief, e.g., chemicals are believed to be dangerous.

MihaiSurdeanu:
I agree with the idea of printing each mention type to a different tsv!

kwalcock:
I added a new printer type, DynamicArgsTsvPrinter, and then tried it out in TestBeliefReader.  The output is in the attached files, including a json version for comparison.  (I had to munge the extensions for the upload.)  The printer was added to the list in HabitusReader.  One can pick and choose there.

[BeliefReader-belief.tsv.csv](https://github.com/clulab/habitus/files/8456316/BeliefReader-belief.tsv.csv)
[BeliefReader-belief_beliefTheme.tsv.csv](https://github.com/clulab/habitus/files/8456318/BeliefReader-belief_beliefTheme.tsv.csv)
[BeliefReader-belief_believer.tsv.csv](https://github.com/clulab/habitus/files/8456319/BeliefReader-belief_believer.tsv.csv)
[BeliefReader.json.txt](https://github.com/clulab/habitus/files/8456323/BeliefReader.json.txt)

maxaalexeeva:
@kwalcock, was it intentional that you have the context key repeated in every row?, e.g, 

```
farmers believe that loans are useful . | nofile | location | date | process | crop | fertilizer
Most women think Trump is a bad person | nofile | location | date | process | crop | fertilizer
Daniel doubts that there will be more virulent mutations | nofile | location | date | process | crop | fertilizer

```

kwalcock:
I believe this is because TestBeliefReader doesn't normally add any Context attribute.  I added a dummy one which will be the same for every line.  Otherwise, without the Context, nothing prints.

```scala
    val context = DefaultContext("location", "date", "process", "crop", "fertilizer", 0)
    val contextMentions = mentions.map { mention => mention.withAttachment(context) }
    printer.outputMentions(contextMentions, "nofile")
```
Some of the other tests in test/scala/org/clulab/habitus/printer check that the values would change, so I *think* it's OK.

maxaalexeeva:
@kwalcock okay! I'll just ignore that for now, but in future, beliefs might get their own context and I can adjust for that then.

kwalcock:
IIRC, In one of Mithun's branches he has a custom context as well.  The intention is that it works mostly automatically.  The fields in the Context should be automatically serialized (assuming they remain in a case class and things like that required for the reflection).

PR #105, opened 2022-04-12
printer 

maxaalexeeva:
- added 'context' attachment to beliefs for proper work of printers--currently N/A
- a small fix for printers


kwalcock:
Thanks so much for finding the bug!  Merging now.

kwalcock:
The mentions could definitely be grouped for efficiency as you suggested.  I'll add it if nobody else does it first.

PR #106, opened 2022-04-13
Group the mentions for printing

kwalcock:


kwalcock:
@maxaalexeeva, can you find another bug?  I think this does what you suggested with the grouping of the mentions.

maxaalexeeva:
@kwalcock tried this with the belief reader on a couple of toy docs---seems to have worked! Thank you!

PR #108, opened 2022-04-18
new variable reader tests

maxaalexeeva:


kwalcock:
Would it be possible for you to grep for
```scala
extends FlatSpec with Matchers
```
and replace it with
```scala
extends Test
```
and any required imports if these nearby files are being changed anyway?  Some of them got away some time ago.  Thank you!

maxaalexeeva:
Closing because there are more current related branches. @kwalcock I think somebody (you?) made the adjustment you requested in another branch.

kwalcock:
I made some adjustments, but I don't see that the ignored test for sent21_12 is in main or the file TestVariableReaderEntities.  Perhaps they are no longer needed.

maxaalexeeva:
@kwalcock yeah @IkeKobby has a different branch going on with all the necessary stuff from here and will open a pr soon

PR #109, opened 2022-04-20
Manage logging

kwalcock:


kwalcock:
@hubert10, this should reduce the logging clutter on your screen so that you can better see important things that are happening.

hubert10:
> @hubert10, this should reduce the logging clutter on your screen so that you can better see important things that are happening.

Thank you @kwalcock !

PR #111, opened 2022-04-20
Standardize tests

kwalcock:
Here's the helper class I'd use to turn these tests into data instead of code.  There might need to be a few adjustments, particularly to make them more easily debugged, but that would be a different round.

### DIFF
--- a/src/test/scala/org/clulab/habitus/variables/TestVariableReader.scala
+++ b/src/test/scala/org/clulab/habitus/variables/TestVariableReader.scala
@@ -43,299 +43,178 @@ class TestVariableReader extends Test {
 
   behavior of "VariableReader"
 
-  // the Clu parser breaks on this one, but the SRL works fine!
-  VariableTest(
-    "sent1", "Farmers’ sowing dates ranged from 14 to 31 July for the WS and from 3 to 11 March for the DS.",
-    "recognize range",
-    Seq(
-      ("sowing dates", Seq(("from 3 to 11 March", "XXXX-03-03 -- XXXX-03-11"))),
-      ("sowing dates", Seq(("from 14 to 31 July", "XXXX-07-14 -- XXXX-07-31")))
+  val variableTests = Array(
+    // the Clu parser breaks on this one, but the SRL works fine!
+    VariableTest(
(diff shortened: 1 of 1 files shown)


kwalcock:
I'm not sure yet whether this would do for TestVariableReaderEntities, but the same strategy might be used there: define some kind of test that describes all the needed information and then have the test class execute the code.  It will depend on how regular the tests are, but there could also be several test classes to handle variations.  I'm not sure whether SimpleTest.reason is necessary.  It looks like someone worked hard to think of some explanations, but they might be write-only and maybe it would be easier to skip them.  @maxaalexeeva?

maxaalexeeva:
Hi @kwalcock ,

Thanks! This looks good. A couple questions:

- how would you feel about including an argument for whether or not the test is passing? With a bunch of new test cases coming in, some tests will probably not be passing write away.
- if we do event testing based on arg labels like we discussed in the meeting (e.g., variable text: `the season with no rains`, variable label: `DrySeason`, value: `March-July`, and we want to make sure that the value is attached to the correct label `DrySeason` instead of a less structured text span `the season with no rains`), we could just have an separate test class based on yours? Might make sense to have them in a separate file, maybe. Let me know if you want to do this or want me to do this.

maxaalexeeva:
Tagging Isaac @ikekobby

kwalcock:
@maxaalexeeva, those are very good points.

I'm figuring that the tests will need to be slightly different in each of the files.  I'm doing TestBeliefReader now and there are passing, failing, toDiscuss, etc. tests.  They will likely be done similar to https://github.com/clulab/eidos/blob/c9b3bbbcda95e4532c3fc2794a9fe8e73661a006/src/test/scala/org/clulab/wm/eidos/text/englishGrounding/TestGrounding.scala#L237.  For TestVariableReader they happened to all be passing, so I skipped it.

In that same file everything had the label "Assignment", so it was hard-coded, but it started out as a parameter to VariableTest and might be the same for some new test class.  You are welcome to write that, especially if it sounds fun.  I probably don't know what constitutes a good test and would have to see some first while you could start sooner.  Either way, though.

PR #114, opened 2022-04-21
Label/norm tests

maxaalexeeva:
- Started a test set that checks events for labels and norms
- added a couple of rules to make the sample test pass

### DIFF
--- a/src/main/resources/variables/entities.yml
+++ b/src/main/resources/variables/entities.yml
@@ -59,6 +59,21 @@
   pattern: |
     /(?i)(crop|cultivar|cultivation|seed|sowed|sown|^plant|grown|growing|variety|varieties)/
 
+
+- name: measurement-rule
+  priority: ${rulepriority}
+  label: Value
+  type: token
+  pattern: |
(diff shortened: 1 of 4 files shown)


maxaalexeeva:
@MihaiSurdeanu, this is based on @kwalcock's restructuring of previously made variable tests---those were checking for variable text and both value text and norm. Right now, the tests  in this file are checking for label and norm, but I have not removed the ability to check for text in case decide we want to do that. Is the sample test what you wanted?

maxaalexeeva:
@IkeKobby, fyi

kwalcock (approved):
That looks OK to me.  FWIW some of the extra indexes and tuples are there so that the error messages are more informative.  When tests fail it is more difficult than usual to identify a unique condition that is bad, so some extra context is added.

maxaalexeeva:
Thanks @kwalcock! Gonna merge this, so that @IkeKobby has access to this in the morning gmt. @MihaiSurdeanu, we can adjust in the next pr if that's not what you were suggesting.

PR #116, opened 2022-04-26
Use entire sentence, not just first word, and separate sentences with NL

kwalcock:


MihaiSurdeanu:
Thanks @kwalcock! LGTM.

PR #118, opened 2022-04-28
Adjusted acronym lexicons

hubert10:
Hi Mihai, the adjusted acronym lexicons are added here.
Thanks!
-Hubert


MihaiSurdeanu:
@hubert10: this is great, but it includes a bunch of other changes that probably don't belong in this merge.
@kwalcock: what branch was this created from? Maybe it should be merged in that one first? Or create a new branch from `case-restore`, and add the adjusted lexicon there?

PR #119, opened 2022-04-28
Adjusted acronym lexicons

hubert10:
Hi Mihai, decline the previous PR and consider this one for adjusted acronym lexicons. Thanks!
-Hubert


MihaiSurdeanu:
Thanks @hubert10 !

kwalcock:
By chance, this PR was made as I was looking into a(n oversimplified) way to check automatically for acronyms (and initialisms and such).  It looks like @hubert10 might be doing this all by hand, so I'm pushing up the announcement in case it saves anyone some work.  Right now it is in the branch [kwalcock/capitals](https://github.com/clulab/habitus/tree/kwalcock/capitals) in the file [AcronymReader](https://github.com/clulab/habitus/blob/kwalcock/capitals/src/main/scala/org/clulab/habitus/AcronymReader.scala).  For the one sample file, it concluded that these need to be all caps:

```
LBA	10
CFA	9
F	6
SAED	4
SSF	4
BP	2
CMS	2
```
I noticed that some terms with initial caps were added to ACRONYM.tsv, so then looked for those and found less definitively
```
Matam	9
Source	7
F	6
Agricole	5
Banque	5
Dagana	5
La	5
Bakel	4
Podor	4
Weekly	4
Guiers	3
Nord	3
Fax	2
Hiv	2
Lac	2
Lake	2
North	2
October	2
Okra	2
Organizations	2
Others	2
Peanut	2
Prev	2
River	2
Rosso	2
Senegal	2
Sup	2
Tel	2
Total	2
```

It can run on a single file like this, but it would be more accurate if run on a larger corpus and the results combined.  The basic idea is to look through sentences for places where words appear in an environment in which their native or underlying case convention is exposed.  For now that is not the first word of a sentence and not the first part of a sentence that includes a run of capitalized words that could be a header or title.  (So, this doesn't include newspaper/book capitalization in the middle of a sentence, or sentences embedded in quotes, or lowercase prepositions/determiners in the middle of a header that would end it, etc.)  If that capitalization is not contradicted somewhere else within the document/corpus and there is enough evidence (right now 2 instances), then it is considered an acronym.  Eventually some case restoration program should leave the word alone and not, for example, turn SAED to Saed.

PR #128, opened 2022-05-05
Conditionally restore case of some sentences, not others

kwalcock:


kwalcock:
...continued from #125.  Here are preserved and conditionally restored versions of the sample sentences attached.

[mix.txt.restored.txt](https://github.com/clulab/habitus/files/8634629/mix.txt.restored.txt)
[mix.txt.preserved.txt](https://github.com/clulab/habitus/files/8634630/mix.txt.preserved.txt)

MihaiSurdeanu:
Looks pretty good!

PR #130, opened 2022-05-09
Processing multiple files

hubert10:
Hi Keith, I added ConditionalCaseMultipleFileApp and LexiconNerMultipleFile for processing and extracting NER for multiple file respectively, please review this PR. Thanks --Hubert


kwalcock:
Thanks, @hubert10.  It looks good to me.  If this eventually becomes very popular, perhaps the duplicate code could be reduced, but for now I think it's plenty good.  Merging now.  Generally that entails removing the branch on the server.  You would usually still have a copy for an emergency.

kwalcock:
If I understand it correctly, the original LexiconNer creates a Source and reads lines from it.  The Source appears never to get closed.  That's not a tragedy if it's only one file, but if there is a large directory of many files as could be the case with LexiconNerMultipleFile, it would become a problem.  I'm going to do another pass over the code and close each Source.

hubert10:
Thanks Keith!

PR #131, opened 2022-05-09
Make sure Source is closed

kwalcock:


MihaiSurdeanu:
Nice!

kwalcock:
One thing that I should document is a change from
```scala
  for(line <- io.Source.fromFile(new File(inputFileName)).getLines()) {
```
to
```scala
      Sourcer.sourceFromFile(inputFile).autoClose { source =>
```
The `Source.fromFile` uses an implicit codec.  I believe to have tracked that down in the past and found that it can be influenced by environment variables such as $LANG and -Dfile.encoding so that it can work differently on different computers and especially for Windows.  Our utility version, Sourcer.sourceFromFile sets the codec explicitly to UTF-8.  It might be wrong, but it should be consistent.  If we're getting files that aren't UTF-8, then this could cause a problem and hopefully someone will run into this comment.

PR #134, opened 2022-05-09
Isaac label test

IkeKobby:
- added a method for running entity tests and added some entity tests
    - some property rules for entity and events tests based on current annotations.


kwalcock:
The CropAnalysis file needs
```scala
import org.clulab.utils.Closer.AutoCloser
```
at the top.  IntelliJ is not good with copy/paste of implicit values.  It misses this one.

IkeKobby:
Okay.

MihaiSurdeanu:
Thank you all! LGTM!

src/main/scala/org/clulab/habitus/variables/CropAnalysis.scala
```diff
@@ -0,0 +1,63 @@
...
+      pw.println("filename\tmention type\tfound by\tsentence\tmention text\targs in all next columns (argType: argText)")
+      for (file <- parFiles) {
+        try {
+          val unfiltered = FileUtils.getTextFromFile(file)
+          val text = cleanText(unfiltered)
+          val filename = StringUtils.afterLast(file.getName, '/')
+          println(s"going to parse input file: $filename")
+          val parsingResults = processor.parse(text)
+          val targetMentions = parsingResults.allMentions
+          val contentMentions = targetMentions.filter(m => m.label == "CropAssignment" || m.label == "Crop")
```
kwalcock, line 41: Is the move of `(!m.isInstanceOf[TextBoundMention]) {` to down below intentional?  It would change the printing slightly.  There is a TsvWriter class that one could use, but it isn't especially important.

kwalcock (commented):
LGTM as well.  Just wanted to double check this.

src/main/scala/org/clulab/habitus/variables/CropAnalysis.scala
kwalcock: However, don't let my comment slow anyone down.  I'm fine to merge.

IkeKobby: Yes  please, it is intentional @kwalcock. Also I had this file here to work on the crop extraction analysis Mihai asked me to do.

PR #136, opened 2022-05-18
redudancy action adjustment

maxaalexeeva:
Avoid applying redundancy action to property assignments because same property can apply to multiple variables. Example:

Farmers used four indica cultivars: Aiwu (short duration, slender grain), I Kong Pao (short duration, bold grain) and IR1529 and Jaya (medium duration, slender grain).

Medium duration and slender grain will attach to both IR1529 and Jaya. 

Note: a better fix will need to happen eventually when all mention types are solidified.

### DIFF
--- a/src/main/resources/lexicons/CROP.tsv
+++ b/src/main/resources/lexicons/CROP.tsv
@@ -178,5 +178,5 @@ sorghum
 peanut
 corn
 vegetables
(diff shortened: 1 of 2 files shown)


IkeKobby:
Thanks @maxaalexeeva .  Merged and pulled.

PR #137, opened 2022-05-24
Combine var readers + context extractor bug fix

maxaalexeeva:
- moved planting area rules to regular variable rules
- removed separate planting reader 
- removed path to grammar in the shell bc now there's only one 
- moved planting area tests to variable tests (label-based tests)
- fixed a couple of context extractor bugs + added a few context extractor tests


maxaalexeeva:
@IkeKobby fyi

src/main/resources/variables/entities.yml
```diff
@@ -40,6 +40,13 @@
   pattern: |
     [lemma = /plant|sow|seed/  & tag = /^VB/] 
 
+- name: variable-area
+  priority: ${rulepriority}
+  label: Area
+  type: token
+  pattern: |
+    ([lemma=/cultivate|plant|sown/])? [lemma=/area/] ([lemma=/sown/])?
```
kwalcock, line 48: I don't know anything about this, but it seems strange that "sown" would be a lemma.  Wouldn't it be "sow"?

src/test/scala/org/clulab/habitus/variables/TestDefaultContextAttachment.scala
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  behavior of "VariableReader Entities"
+
+  val variableTests: Array[DefaultContextAttachmentTest] = Array(
+    DefaultContextAttachmentTest(
+      "sent-1",
+      "The area sown with rice and fertilized with urea for this 2021/2022 wintering campaign is 28,223 ha in wintering in Senegal.",
+      "PlantingArea",
+      Map(
```
kwalcock, line 58: I like this use of a Map for this small number of tests.  If there were to be lots of them eventually, then the keys might be turned into constants/variables so that the compiler will check for typos or in case the name ever changes.  val DATE = "date".  If they all have the same keys, it could be a class instead, possibly with automatic defaults of missing values to "N/A".

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
```diff
@@ -96,7 +97,7 @@ trait ContextExtractor {
       val contextSentences = getSentIDsInSpan(mention, windowSize)
       val instances = getInstancesInContext(contextType, contextSentences, allMentions)
       if (instances.nonEmpty) {
-        return instances.groupBy(identity).map(i => i._1 -> i._2.length).max._1.toLowerCase()
+        return instances.groupBy(identity).map(i => i._1 -> i._2.length).maxBy(_._2)._1
```
kwalcock, line 100: It doesn't matter, but if one wanted to get rid of this return, one might use recursion on the windowSize.

kwalcock (approved):
LGTM if that sown is OK.

src/main/resources/variables/entities.yml
maxaalexeeva: Yes, absolutely should be sow. Thank you!

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
maxaalexeeva: I don't think I know how to do that :'(

src/test/scala/org/clulab/habitus/variables/TestDefaultContextAttachment.scala
maxaalexeeva: I did not completely understand about the class, but is what i did in this commit what you meant about variables? https://github.com/clulab/habitus/pull/137/commits/5277b4df279050e1dd7b3d119971183c8bcf44b5

kwalcock: Yes.  If it was me, I would have made a typo in one of the keys already, but now the compiler will cover me.  The other way is something like

```scala
  case class Record(location: String = NA, date: String = NA, process: String = NA, crop: String = NA, fertilizer: String = NA, comparative: String = NA)

  Record(location = "Senegal", date = "2021/2022", process = "planting", comparative = "0")
```
which guarantees that the set of keys is fixed so that I can't add "oops" -> "a daisy" and the default value is automatic, .  However, it was mostly a thought experiment and I wouldn't change what's there without a better reason.

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
kwalcock: It's not important.  Some people have taken an oath never to use "return" and have thought about ways to avoid it.  If it changes some day, I'll tag you.

src/test/scala/org/clulab/habitus/variables/TestDefaultContextAttachment.scala
maxaalexeeva: I see! That does look nice...

maxaalexeeva: I know you said not to change it... But I changed it... It is just so much more beautiful this way!

kwalcock: I have to learn to keep my mouth closed ;-)

maxaalexeeva: Nooo! I need to learn!! Thanks!!

maxaalexeeva: Does it look ok now? Should I merge?

kwalcock: I was just waiting on Mihai since he was named as another reviewer and I wouldn't want to cross him.  Something in a different email hinted at a disagreement.  It's fine with me.

maxaalexeeva:
@MihaiSurdeanu would you have a look at this pr? look ok to merge? Thanks!

MihaiSurdeanu:
LGTM! Thanks for working on it and for being patient!

kwalcock:
Merged!

maxaalexeeva:
@MihaiSurdeanu thanks!

maxaalexeeva:
@kwalcock thank you!

PR #138, opened 2022-06-02
procVer update + outputting label

maxaalexeeva:
- updated procVer and adjusted tests
- adding label to tsv output (there are multiple types of mentions extracted in the same reader, so we need to be able to differentiate between them)


kwalcock:
It looks mighty fine to me.  If those numerous tests that had to be adjusted get too cumbersome, they can be rethought.  Tests sometimes impede progress.

PR #140, opened 2022-06-07
Remove sentences with too many single-character words

kwalcock:


kwalcock:
I checked tika and scienceparse and they both want to generate lots of single-letter words for FCR_SRVRice.pdf.  In this case, the Adobe converter seems to do significantly better.  One can compare the files below.  I added to the bad sentence check in habitus, but I suppose that the PDF preprocessor could have done this check as well.

[FCR_SRVRice.scienceparse.txt](https://github.com/clulab/habitus/files/8857236/FCR_SRVRice.scienceparse.txt)
[FCR_SRVRice.adobe.txt](https://github.com/clulab/habitus/files/8857237/FCR_SRVRice.adobe.txt)

kwalcock:
I shouldn't merge my own code without more eyes looking at it first, so @IkeKobby, is this a reasonable way to deal with those strange strings of characters?  It's unfortunate that the rest of the sentence gets sacrificed.

IkeKobby:
Hi @kwalcock , in `[FCR_SRVRice.scienceparse.txt](https://github.com/clulab/habitus/files/8857236/FCR_SRVRice.scienceparse.txt)` there are still some sentence with the spaced-out characters. 

But `[FCR_SRVRice.adobe.txt](https://github.com/clulab/habitus/files/8857237/FCR_SRVRice.adobe.txt)` looks good.

kwalcock:
@IkeKobby, the sentences will still be in the txt file, but they will not be processed by habitus.  They will be skipped like the too long and too short sentences.

PR #141, opened 2022-06-09
Binarizing events + unit appropriateness check

maxaalexeeva:
- added an action for splitting events with multiple non-complementary values into binary events
- added an action for checking if the unit is appropriate for the label 
- added 'yield' to lemmas that indicate harvest
- modified tests to account for the changes
- removed an unused file


src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
```diff
@@ -170,5 +170,45 @@ class HabitusActions extends Actions {
...
+    list.contains(string)
+  }
+
+  def measurementIsAppropriate(m: Mention): Boolean = {
+    // check if any of the possible units shows up in the norm
+    val probablyUnit = m.norms.get.head.split("\\s").last // e.g., get "m2" from "6 m2"; assume value is separated from unit with a space and unit is one token
+    val appropriateUnits = m.label match {
+      case "Quantity" => Seq("t/ha", "kg/ha", "kg", "d", "cm") // fixme: added d and cm for now to make sure tests pass, but do we want days and cm to be quantities?
+      case "AreaSize" => Seq("ha")
+      case _ => ???
```
kwalcock, line 201: It's trivial, but could a runtime exception be thrown here with a message, like "Unknown measurement label ${m.label}"?  Right now it is clear from the rules that it can never happen, but later it might not be and it would be handy to see the culprit label which isn't displayed for a NotImplementedError thrown by ???.  Someone could fix it then without needing to add a println or start a debugger.  Some places probably have a lint rule for that.

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
```diff
@@ -170,5 +170,45 @@ class HabitusActions extends Actions {
...
+
+  def listStringOverlap(string: String, list: Seq[String]): Boolean = {
+    list.contains(string)
+  }
+
+  def measurementIsAppropriate(m: Mention): Boolean = {
+    // check if any of the possible units shows up in the norm
+    val probablyUnit = m.norms.get.head.split("\\s").last // e.g., get "m2" from "6 m2"; assume value is separated from unit with a space and unit is one token
+    val appropriateUnits = m.label match {
+      case "Quantity" => Seq("t/ha", "kg/ha", "kg", "d", "cm") // fixme: added d and cm for now to make sure tests pass, but do we want days and cm to be quantities?
```
kwalcock, line 199: This Seq and the above listStringOverlap make me wonder if Set is the way to go because it has an apply method returning a Boolean for membership as if it was intended for this.  For example,
```scala
val quantityUnits = Set("t/ha", "kg/ha", "kg", "d", "cm")
val isAppropriate = quantityUnits(probablyUnit)
```
One could go overboard with something like
```scala
// This can be static in the companion object.  
val labelToAppropriateUnits = Map(
  "Quantity" -> Set("t/ha", "kg/ha", "kg", "d", "cm"),
  "AreaSize" -> Set("ha")
)
// In the method after probableUnit is extracted...
labelToAppropriateUnits
    .getOrElse(m.label, throw new RuntimeException(s"Unknown measurement label ${m.label}"))(probableUnit)
```

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
```diff
@@ -170,5 +170,45 @@ class HabitusActions extends Actions {
...
+    }
+    )
+    for (m <- target) {
+      for (v <- m.arguments("value")) {
+        val newArgs = Map("variable" -> m.arguments("variable"), "value" -> Seq(v))
+        val newMention = copyWithArgs(m, newArgs = newArgs)
+        toReturn.append(newMention)
+      }
+    }
+    toReturn ++ other
```
kwalcock, line 188: I see an ArrayBuffer, which always puts me on edge.  You might consider something like this which Scala does surprisingly nicely.

```scala
    val splitTargets = for {
      m <- targets
      value <- m.arguments("value")
      newArgs = Map("variable" -> m.arguments("variable"), "value" -> Seq(value))
    } yield copyWithArgs(m, newArgs)
    splitTargets ++ other
```

kwalcock (commented):
Excellent (in spite of my interference)!

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
```diff
@@ -170,5 +170,45 @@ class HabitusActions extends Actions {
...
+      case "Quantity" => Seq("t/ha", "kg/ha", "kg", "d", "cm") // fixme: added d and cm for now to make sure tests pass, but do we want days and cm to be quantities?
+      case "AreaSize" => Seq("ha")
+      case _ => ???
+    }
+    listStringOverlap(probablyUnit, appropriateUnits)
+  }
+
+  def appropriateMeasurement(mentions: Seq[Mention]): Seq[Mention] = {
+    // applies to individual rules to check if the measurement is appropriate for the mention label
+    for {
```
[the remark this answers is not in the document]
maxaalexeeva: That's beautiful!

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
maxaalexeeva: This looks good, but I am not sure I understand where the labelToAppropriateUnits map should go.

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
maxaalexeeva: Yeah, I remembered to use for/yield in one of today's methods based on your previous comments regarding array buffers, but did not think of it here. To be honest, I hate these for/yield things because I can't print out inside of them :D But I am sure I will phase out my love for array buffers soon :D

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
maxaalexeeva: @kwalcock I have changed things like you suggested except for the placement of the label/unit map.

kwalcock:
Does Mihai need to check this?  If not, I'm ready to merge.

If that `val labelToAppropriateUnits = Map(` could be moved just above `def measurementIsAppropriate(m: Mention): Boolean = {`, then it would still be in the immediate vicinity and I'd be sure that it isn't re-created on each method call.  I should figure out if Java can cache that somehow.  Other languages support `static` right then and there to make it obvious.

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
kwalcock: That's a very good point about the printing, so good that I looked it up once.  Will this work for you?

```scala
    val splitTargets = for {
      m <- targets
      _ = println(m)
      value <- m.arguments("value")
      _ = println(value)
      newArgs = Map("variable" -> m.arguments("variable"), "value" -> Seq(value))
      _ = println(newArgs)
    } yield copyWithArgs(m, newArgs)
```

maxaalexeeva:
@kwalcock let's check! (and thanks for all the feedback!)
@MihaiSurdeanu, all the changes should be close to what we previously discussed. I want @IkeKobby to have access to these changes asap before he starts to fix other issues we found, but I can wait till morning to merge in case you have comments. If you have comments after I merge, we can do another pr to address those. Hope that's ok.

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
maxaalexeeva: Neat! Did not know! Thanks!

MihaiSurdeanu:
Looks great @maxaalexeeva ! Ok to merge.
I wonder if we should to talk to @hubert10 to see what valid units we should add in that action.

maxaalexeeva:
@MihaiSurdeanu thanks! Yes, good idea.

PR #145, opened 2022-06-22
some rule and test adjustments

maxaalexeeva:
- adjusted var reader tests to check text and not label
- added shouldable to var reader and var reader labeled 
- some small rule and test adjustments


IkeKobby (approved):
Thank you @maxaalexeeva. Looks good.

PR #146, opened 2022-06-23
Test for custom kerning

kwalcock:
Also use ratio for non-alpha limit


kwalcock:
The changes may have taken out more sentences than intended.  I'll check.

kwalcock:
@IkeKobby, here's the code that goes along with the email, at least the code so far.

kwalcock:
On second thought, I'm not sure this is a good idea.  I don't like that some of the processing has to be disabled for the tests that include legitimate sentences like "Seeding occurred on October 1, 2019."  This has half tokens that are non-alpha: 1 comma 2019 period.  Maybe the non-alpha test needs to be changed.  It could instead count the number of non-alpha words in a row and keep that within a reasonable range.

kwalcock:
The new test was changed to check for (floating point) numbers in a row.  The example sentence had lots of them.

PR #154, opened 2022-07-15
Fixing var reader issues

IkeKobby:
- Added tests
- Separated entity tests into entity vs. norms
- Added various rules and test adjustments.


src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
```diff
@@ -186,20 +207,95 @@ class HabitusActions extends Actions {
...
-    "Quantity" -> Set("t/ha", "kg/ha", "kg", "d", "cm"),
-    "AreaSize" -> Set("ha")
+    "Quantity" -> Set("t/ha", "kg/ha", "kg", "d", "cm", "mg/l", "kg n ha-1"),
+    "AreaSize" -> Set("ha"),
+    "YieldAmount" -> Set("t/ha", "kg/ha", "kg"),
+    "FertilizerQuantity" -> Set("kg/ha", "mg/l", "kg n ha-1")
   )
 
+  def hasLetters(string: String): Boolean = {
+    string.exists(ch => ch.isLetter || ch.toString == "/")
```
kwalcock, line 217: Is that the same as ch == '/' ?

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
```diff
@@ -170,8 +172,27 @@ class HabitusActions extends Actions {
...
+  }
+
+  def fertilizerQuantityActionFlow(mentions: Seq[Mention]): Seq[Mention] = {
+//    splitIntoBinary(mentions).filter(m => allowableTokenDistanceBetweenVarAndValue(m, 12))
+    appropriateMeasurement(splitIntoBinary(mentions).filter(m => allowableTokenDistanceBetweenVarAndValue(m, 12)))
+  }
+
+  def allowableTokenDistanceBetweenVarAndValue(mention: Mention, maxDist: Int): Boolean = {
+    val variable = mention.arguments("variable").head
+    val value = mention.arguments("value").head
```
kwalcock, line 188: I count 14 uses of "value" and 8 of "variable".  I think that's sufficient to turn the string into a variable and reference it.  The compiler can check that somebody (me) has not misspelled it.  Something like
```scala
val VALUE = "value"
mention.arguments(VALUE).head
```

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
```diff
@@ -186,20 +207,95 @@ class HabitusActions extends Actions {
...
+    )
+    newEvent
+  }
+
+  def adjustQuantityNorm(mentions: Seq[Mention]): Seq[Mention] = {
+    mentions.map { m =>
+      val value1 = m.arguments("value1").head.text.toFloat
+      val value2 = m.arguments("value2").head.text.toFloat
+      val unit = m.arguments("unit")
+      val valueRange = s"${value1} -- ${value2}"
```
kwalcock, line 281: There is a danger here that you won't get the expected string.  For example,
```scala
scala> val f1 = "0.0001".toFloat
f1: Float = 1.0E-4

scala> val f1 = "33333.333".toFloat
f1: Float = 33333.332
```

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
IkeKobby: Hi Keith, technically yes. I want to double check and address.

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
IkeKobby: Okay. That’s good.

src/main/scala/org/clulab/habitus/utils/ContextExtractor.scala
```diff
@@ -46,9 +48,13 @@ trait ContextExtractor {
       "harvesting"
     } else if (lemmas.exists(l => creditLemmas.contains(l))) {
       "credit"
+    } else if (lemmas.exists(l => irrigationLemmas.contains(l))) {
+      "irrigation"
+    } else if (lemmas.exists(l => weedsLemmas.contains(l))) {
+      "weeds"
```
kwalcock, line 54: I think this is long enough to warrant something like this:
```scala
  case class ProcessToLemmas(process: String, lemmas: Set[String]) {
    def this(processAndLemmas: (String, Set[String])) = this(processAndLemmas._1, processAndLemmas._2)
  }

  // These are prioritized highest to lowest because there can be multiple matches.
  val processToLemmasMap = Seq(
    "planting"         -> Set("plant", "sow", "cultivate", "cultivation", "grow"),
    "harvest"          -> Set("harvest", "yield"),
    "credit"           -> Set("credit", "finance", "value"),
    "irrigation"       -> Set("irrigation", "irrigate"),
    "weeds"            -> Set("weed"),
    "natural_disaster" -> Set("flood", "bird", "attack")
  ).map(new ProcessToLemmas(_)) // just for convenience

  def getProcess(mention: Mention): String = {
    val lemmas = mention.sentenceObj.lemmas.get
    val process = processToLemmasMap
        .find { processToLemmas =>
          lemmas.exists(processToLemmas.lemmas)
        }
        .map(_.process)
        .getOrElse("UNK")

    process
  }
```

src/main/scala/org/clulab/habitus/variables/VariableProcessor.scala
```diff
@@ -36,8 +37,7 @@ class VariableProcessor(val processor: CluProcessor,
   def parse(doc: Document): ParsingResult = {
     // extract mentions from annotated document
     val mentions = extractor.extractFrom(doc).sortBy(m => (m.sentence, m.getClass.getSimpleName))
-
-    val (tbms, relsAndEvents) = mentions.partition(_.isInstanceOf[TextBoundMention])
+    val (tbms, _) = mentions.partition(_.isInstanceOf[TextBoundMention])
```
kwalcock, line 40: Since the second partition is being thrown away, is this
`val tbms = mentions.filter(_.isInstanceOf[TextBoundMention])`
or
`val tbms = mentions.collect(mention: TextBoundMention => mention }`

src/main/scala/org/clulab/habitus/utils/DefaultContextExtractor.scala
```diff
@@ -16,7 +16,7 @@ class DefaultContextExtractor extends ContextExtractor {
 
       val thisSentMentions = mentionsBySentence(i).distinct
       val (thisSentTBMs, thisSentEvents) = thisSentMentions.partition(_.isInstanceOf[TextBoundMention])
-      val thisSentDates = thisSentTBMs.filter(_.label == "Date")
+      val thisSentDates = thisSentTBMs.filter(m => m.label == "Date" || m.label == "DateRange")
```
kwalcock, line 19: So even though the label is "DateRange" here, getContext will be called with "Date".  Because of the duplication of the strings and the checking of the string in getContext(), I'll suggest a minor design change next time.  Let's first see whether the code survives.

IkeKobby: Okay. Sounds good.

kwalcock (commented):
Thanks for all the hard work on this.  It looks quite good.

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
IkeKobby: Keith please do you have any suggestion for this?

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
IkeKobby: At the moment changing `ch == '/'` does not work.

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
kwalcock: It might be even worse, but one could use the original text: `m.arguments("value1").head.text`.  If someone wrote "3.14", "03.14", "0.314E1", "+3.14" or "--3.1.4" it might be a problem that there is more than one representation of the same value.  To avoid the rounding error one could start with .toDouble or (better) use a BigDecimal, and to keep from switching to scientific notation one can probably use f"" with some kind of formatting instead of s"".  (https://docs.scala-lang.org/overviews/core/string-interpolation.html).  I don't know how important it is to account for this, but did want to flag potential problems.

kwalcock: I'd be perfectly happy if a comment was added near those lines so that the next person or our future selves are alerted to the possible gotcha.

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
IkeKobby: yes, it does warn; `comparing values of types Char and String using `==` will always yield false` but having `ch.toString` works without warn.

kwalcock: Do you notice that the quotation marks are now apostrophes?

kwalcock: (It's different from Python.)

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
kwalcock: There may be a few more instances than what you found so far.

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
IkeKobby: yesss, sure. It is corrected. 

Thank you.

src/main/scala/org/clulab/habitus/actions/HabitusActions.scala
```diff
@@ -170,36 +173,131 @@ class HabitusActions extends Actions {
       )
     }
   }
+
+  def yieldAmountActionFlow(mentions: Seq[Mention]): Seq[Mention] = {
+//    splitIntoBinary(mentions).filter(m => allowableTokenDistanceBetweenVarAndValue(m, 16))
```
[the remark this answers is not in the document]
maxaalexeeva: and line  117 under entites.yml (action: appropriateMeasurement)
Thanks!

MihaiSurdeanu:
Hi @maxaalexeeva,
This looks very good! If @kwalcock is also Ok, we can merge this.
One comment: `labelToAppropriateUnits` seems kinda short. Are these all units you saw in the data?
Thanks!

MihaiSurdeanu (approved):

kwalcock:
I'm good with it but will let Masha respond before I push the button.

maxaalexeeva:
@MihaiSurdeanu I don't think we saw any more units/nothing fails because of this map, so I think it's most of them. We can add more if/when we find them.

MihaiSurdeanu:
Then, @kwalcock please merge!

PR #158, opened 2022-07-28
Output pseudo-paragraph

maxaalexeeva:
Adding a +/-n sentence context window column/field to output. Getting to real paragraph looks like it would complicate some methods + paragraphs we get from pdfs are not always reliable and they might be too big (thus overwhelming within a document). Context window size is negotiable (set in application.conf).

### DIFF
--- a/src/main/scala/org/clulab/habitus/printer/MentionInfo.scala
+++ b/src/main/scala/org/clulab/habitus/printer/MentionInfo.scala
@@ -3,15 +3,18 @@ package org.clulab.habitus.printer
 import org.clulab.habitus.utils.Pairable
 import org.clulab.odin.Mention
 
-case class MentionInfo(sentenceText: String, inputFilename: String, label: String) extends Pairable
+case class MentionInfo(contextWindow: String, sentenceText: String, inputFilename: String, label: String) extends Pairable
 
 object MentionInfo {
 
   def apply(mention: Mention, inputFilename: String): MentionInfo = {
+    val windowSize = 3
     val document = mention.document
+    val sentInterval = (math.max(mention.sentence - windowSize, 0), math.min(mention.sentence + windowSize, mention.document.sentences.length + 1))
+    val contextWindow = mention.document.sentences.slice(sentInterval._1, sentInterval._2).map(_.getSentenceText).mkString(" ")
     val sentenceText = document.sentences(mention.sentence).getSentenceText
     val label = mention.label
-    new MentionInfo(sentenceText, inputFilename, label)
+    new MentionInfo(contextWindow, sentenceText, inputFilename, label)
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/main/scala/org/clulab/habitus/printer/MentionInfo.scala
```diff
@@ -1,17 +1,24 @@
...
+case class MentionInfo(contextWindow: String, sentenceText: String, inputFilename: String, label: String) extends Pairable
 
 object MentionInfo {
 
+  val config = ConfigFactory.load()
+  val windowSize: Int = config[Int]("VarDatesReader.contextWindowSize")
+
   def apply(mention: Mention, inputFilename: String): MentionInfo = {
     val document = mention.document
+    val sentInterval = (math.max(mention.sentence - windowSize, 0), math.min(mention.sentence + windowSize + 1, mention.document.sentences.length + 1))
```
kwalcock, line 17: I *think* there might be something off on the right side of this interval.  Isn't it usually math.min(..., _.length)?  length is typically already one beyond the indexable values and adding another would be out of bounds.

maxaalexeeva: Ah right thanks!

kwalcock: Around this line, `mention.document` could just be `document`, but see the next comment.  One could drill down directly to the sentences.

src/main/scala/org/clulab/habitus/printer/MentionInfo.scala
```diff
@@ -1,17 +1,24 @@
...
+    val contextWindow = mention.document.sentences.slice(sentInterval._1, sentInterval._2).map(_.getSentenceText).mkString(" ")
```
kwalcock, line 18: If that sentInterval was not a Tuple2[Int, Int] but rather a Range, then quite a bit of math could be removed and it wouldn't be necessary to program by the numbers with _._1 and _._2, which can be difficult to understand.  Next time I have to do this myself, I will probably write something like
```scala
import org.clulab.struct.Interval

val sentences = mention.document.sentences
val mentionRange = Interval(mention.sentence).indices // for lack of a Range(Int) constructor :-(
val sentRange = Range(mentionRange.start - windowSize, mentionRange.end + windowSize).intersect(sentences.indices)
val contextWindow = sentRange.map(sentences(_).getSentenceText).mkString(" ")
```

kwalcock (approved):
I learned something about Ranges :-)

src/main/scala/org/clulab/habitus/printer/MentionInfo.scala
maxaalexeeva: @kwalcock, that looks great! I especially like the intersect there. Updated.

maxaalexeeva:
@kwalcock oops, thanks, forgot to change the document part

kwalcock:
I can be very pedantic.  It's Mihai's turn now.

MihaiSurdeanu:
LGTM!

PR #162, opened 2022-09-01
Update processors

kwalcock:


kwalcock:
@maxaalexeeva, this might conflict slightly with your branches that already have the change.  The processors version you have probably needs to lose the -SNAPSHOT.

maxaalexeeva:
@kwalcock thank you!!! I decided to postpone relabeling the tests that were supposed to pass with new processors until my latest branch is done---there's way too much stuff that I changed I think and those fixedWithNewProcRelease tests might be impacted.

PR #163, opened 2022-09-19
added more-crops-lexicons

hubert10:
Hi @maxaalexeeva, I have added about 800 lexicons and the list is still long, I will add more later. Thanks


maxaalexeeva:
Thank you, @hubert10! That's great! I am wondering if there is an indication of what those are varieties of? Like there were a bunch of peanut varieties that I found (e.g., Fleur 11). Or I googled `LAC 23` from the varieties you added, and it looks like it's a rice variety. If we know which variety goes with which crop, we could use crop-specific variety lexicons. That might help with efficiency of lexicon ner maybe, @MihaiSurdeanu?

hubert10:
I added Rice, Maize and Sorghum. The remaining are Millet, Peanut, Cowpea,
Cassava, Yam, Potato, Tomato and Onion.

kwalcock:
I agree that it would help to organize these a little, perhaps by adding a comment (if that works) like

# The following are varieties of rice:

or by grouping them into separate files, and in either case by alphabetizing the entries within each group and subgroup.  That makes it more obvious when there are duplicates and allows one to notice quickly if there are clusters around a prefix.   If I need to add some comment processing/ignoring code somewhere, just let me know.

kwalcock:
P.S.  I checked and believe that comments are initiated with `//` so the above statement was fixed.

hubert10:
Any of the above suggestions is fine with me, so @keith you prefer comments
rather than separate files?

kwalcock:
I have no preference, but I *think* that the file name is used for the named entity tag so that if you have multiple files you will get multiple tags.  Instead of CROP, you might have RICE and MAIZE which would cause other problems.

maxaalexeeva:
I didn't think about how the file name tells us the entity name. Good point. I think with splitting varieties into separate files, I was thinking of something like "if we are working with peanuts, use this generic crop.tsv + this peanut_variety.tsv as crop.tsv" (like append the second file to the first one and use that combo as crop lexicon). Same for if it's rice: use generic crop + rice-specific varieties file. I don't really know if that would help with anything, to be honest. Would it help with efficiency in any way, @kwalcock? Or is organizing crop varieties in some way mainly for esthetic purposes?

kwalcock:
@maxaalexeeva. I'm not sure what is best.  The label comes from the file name and `CROP` seems to be used in two places right now: `entities.yml` and `EntityHistogramExtractor.scala`.  It would be possible, but not handy, to write some code that collects the contents of multiple files at runtime and unifies them under one name.  Another possibility is what I think you suggest, combining the files during development.  CROP.tsv = generic_crop.tsv + peanut_variety.tsv.  The better option might depend on how often and under what circumstances crops are changed.  Would different CROP.tsv files need to live in different code branches or could they be put into different subdirectories so that we have Senegal/CROPS.tsv and Nepal/CROPS.tsv?  If we throw all crops together, how much of a problem will it cause?  Will it be unnecessarily slow or result in false positives or be problematic if/when names of crop varieties overlap?

hubert10:
@maxaalexeeva and @kwalcock, I have added all the lexicons with comment like `// The following are varieties of rice:
` please review this PR. Before pushing to my branch, I pulled main and merged it back with my branch and I suspect that CI checking is failing please review. Thanks

kwalcock:
Thanks @hubert10.  That satisfies my concern about the information being available somewhere.  It can always be rearranged later if need be.  Now I know what a cowpea is.  I'm looking into the failing test.

kwalcock:
The error shows up for this test:

```scala
    VariableTest(
      passingTest,
      "sent10",
      "The Senegal River Valley, located in Sahel zone, is one of the major irrigated rice-producing areas in Senegal.",
      Seq(
        "Location" -> Seq("Senegal", "Sahel", "Senegal River Valley"),
      )
    ),
```
and the problem seems to be that Sahel can now be a potato variety.  If I comment that out in CROP.tsv, then the test passes.  Something  must have changed in the main branch between when you made your branch and when main was merged back, and that change in combination with the new potato variety does not work.  It might be good that it showed up because there probably needs to be a contingency plan for when a variety and a location overlap.  Can @maxaalexeeva and you figure out what to do about it?

maxaalexeeva:
@kwalcock and @hubert10, I have checked a couple of obvious places, but haven't found the issue yet. I am working on a pr and  will keep my eyes open for this.

maxaalexeeva:
Oh, wait, so what happens is the entity is now B-CROP because of the lexicon NER, so of course the location rule does not work... @kwalcock

maxaalexeeva:
@kwalcock, @hubert10, my inclination right now is to just exclude Sahel from the crop lexicon. All rice Sahel varieties were more specific, e.g., Sahel-108. And we do not read for potatoes now. I don't want to exclude a prominent location because of a variety that might not come up. Maybe we can comment it out for now with a note about this?

kwalcock:
That seems OK to me.  On a checklist somewhere we might note the need to compare typical locations in a new country to any existing crop varieties to ensure there is no problematic overlap.  An overly complicated way to do it would be to use a custom NER knowledge base containing local locations to override any crops present in what are called the standard knowledge bases that might get into the way.  One could also disallow overwriting of the location tag with a crop tag in some kind of custom named entity recognizer.  Both those seem excessive right now.  I'll push a tiny change with that line commented out.

kwalcock:
If nobody objects tomorrow, I'll merge this.  Thanks again, @hubert10, for tracking down the varieties.

hubert10:
@kwalcock, ohh thanks for commenting out this Sahel variety, indeed it is a potato variety.

PR #167, opened 2022-09-29
Demonstrate  adding year to document

kwalcock:


kwalcock:
@maxaalexeeva, this is meant to show how one might work with a publication year.  The year should be available from the document or any mention that includes a reference to the document via `YearDocumentAttachment.getYear(doc).get`.

PR #168, opened 2022-10-04
Take a close look at some Sets and Maps

kwalcock:


kwalcock:
There's possibly one more unreliable test.

kwalcock:
@maxaalexeeva, just to keep you in the loop, especially about DefaultContextExtractor.scala.

maxaalexeeva:
@kwalcock thanks for this! I've been dreading making context extractor more presentable! :embarrassed_emoji_goes_here:.
I think the issue with the failing test is that there are two values there, so the label map doesn't work and instead of using the data found within the mention itself, the date is picked with findClosestNotOverlapping within the getContext method. Here's what the mention looks like:

```
List(PlantingEvent, Event) => sowed Jaya between 20 June and 1 July
        ------------------------------
        Rule => planting-event-cop-2
        Type => RelationMention
        ------------------------------
        variable (Planting, Entity) => sowed
        value (Crop, Entity) => Jaya
        value (DateRange, Date, Entity) => between 20 June and 1 July

```
Another possible issue is that there are Dates and DateRanges. I think I handled it in some awkward way in the past, but the new code might not?

maxaalexeeva:
@kwalcock good point. That's just my oversight. I am thinking if they overlap, distance could either be the distance between starts of the mention or the distance could be 0. Zero probably makes most sense?..

kwalcock:
@maxaalexeeva, you probably noticed that I lost track of your

```scala
if (menArgLabels.exists(_._1 contains "Date")) menArgLabels.filter(_._1 contains "Date").head._2 else  getContext(m, "Date", thisSentDates, mentions),
```
which will work on both Date *and* DateRange and replaced it with code that only works on Date.  That was my mistake that your subsequent removal of DateRange fixed.  Thanks!  I'm going to add a commit that corrects the oversight for the record and then remove it again because it isn't needed.

maxaalexeeva:
@kwalcock sounds good! Thank you!

kwalcock:
This was not an unreliable test then, but it shows what problems their existence elsewhere can cause.

maxaalexeeva:
@kwalcock for sure!

kwalcock:
The changes that I made weren't supposed to change behavior and didn't really need to be merged, but the removal of DateRange is significant, so I plan to merge this shortly and continue checking for unreliable behavior in a different branch.  FYI, world.

maxaalexeeva:
Hi @kwalcock! Would you like to merge this?

kwalcock:
Thanks for the reminder!

PR #169, opened 2022-10-05
getting rid of DateRange

maxaalexeeva:
@kwalcock this is in case you think it's reasonable to get rid of date range - this does fix the test that's failing in your branch and also, having both date and dateRange has caused me issues in the past when handling context


kwalcock:
That's plenty fine with me.  I'll do the hard work of pressing the button.

PR #170, opened 2022-10-05
add country to output

maxaalexeeva:
- added country to output based on region-country code lexicon (lexicons for individual countries can be found here https://download.geonames.org/export/dump/); the country itself needs to be added to the lexicon as well. 
- added publication year


maxaalexeeva:
Hi @kwalcock, I have some questions:
1) is there a way to pass a relative path in config [here](https://github.com/clulab/habitus/blob/98bd54e4632d116bc7143d77b33880d6c9afab64/src/main/resources/application.conf#L9)?
2) should I add a notebook/script to process the regions files somewhere in this repo?
3) for the code to work correctly, the country itself needs to be added to the regions lexicon; where in the repo do you think I should add instructions on that?

Thanks!

p.s., also this pr will clash with your maps inspect pr, but I can fix conflicts after you have merged that one.

kwalcock:
1. Yes.  The value can begin with ./ or ../.  You have to watch carefully what the current directory is in relation to the lexicon.  ~ does not work, but something like {user.dir} can be made to work and I can track that down if necessary.  IntelliJ and sbt sometimes have a different opinion on the current directory, especially for subprojects.
2. I don't really know what's up with the regions, but if some raw data needs to be processed with code, we should probably save the code somewhere and document where the resource came from.  I'd favor plain Python over a Jupyter notebook.
3. Documentation is easy to move.  There could be a README.md file in src/python or src/scala or instructions could be a wiki page in github or there could be a special doc directory.

src/main/scala/org/clulab/habitus/utils/DefaultContextExtractor.scala
```diff
@@ -87,14 +97,14 @@ class DefaultContextExtractor extends ContextExtractor {
           require(labels.length == labels.distinct.length, "Labels should be distinct in order to use them in keys of a  map.")
           val menArgLabels = arguments.map(men => men.label -> men.text).toMap
           val location   = getIrrespectiveContext(m, "Location",   thisSentLocs,   None)
+          val country = DefaultContextExtractor.regionMap.getOrElse(location, "N/A")
```
kwalcock, line 100: I don't know if it is feasible or excessive, but I would favor keeping track of these Optional values a little longer as such (i.e., Option[String]) and only turn them into "N/A" (or "" for tsv/csv or null for JSON) until the very last minute.  I would find it difficult to know whether to check for "N/A" unless maybe the variable was called countryOrNA.  There are lots of N/As floating around, so I wouldn't necessary change this one, but rather review them all (at a later date, optionally).

src/main/scala/org/clulab/habitus/utils/DefaultContextExtractor.scala
```diff
@@ -106,4 +116,26 @@ class DefaultContextExtractor extends ContextExtractor {
...
+
+object DefaultContextExtractor {
+
+  val config = ConfigFactory.load()
+  val localConfig: Config = config[Config]("VarReader")
+  val inputDir: String = localConfig[String]("regionsLexicon")
+  val files = FileUtils.findFiles(inputDir, ".tsv")
+  val regionMap = mutable.Map[String, String]()
+  files.foreach { f =>
+    val source = Source.fromFile(f)
```
kwalcock, line 134: Here's where I wonder how this would work from a jar file.  Perhaps that will never happen.

src/test/scala/org/clulab/habitus/variables/TestPrinting.scala
```diff
@@ -10,7 +10,8 @@ import java.io.File
 class TestPrinting extends Test {
   implicit val formats: DefaultFormats.type = org.json4s.DefaultFormats
 
-  val inputDir = "./src/test/resources"
+  val inputDir = "./src/test/resources/txt"
```
kwalcock, line 13: For testing I think it's perfectly fine to rely on the development environment being present.

kwalcock (commented):
Nice work!  As usual I couldn't stop talking, but don't let that worry you :-)

python/processing_geonames/get_essential_region_info_from_geonames.py
```diff
@@ -0,0 +1,33 @@
...
+input_dir = args.input
+output_dir = args.output
+
+if not os.path.exists(output_dir):
+    os.makedirs(output_dir)
+
+def main():
+    files = [f for f in listdir(input_dir) if isfile(join(input_dir, f))]
+    for file in files:
+        data = pd.read_csv(os.path.join(input_dir, file), header=None, sep="\t")
```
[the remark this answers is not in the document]
maxaalexeeva: @kwalcock, in this file, there are two columns: one has special symbols and the other doesn't (very thoughtful of the creators!). I made an executive decision to just use the column with no accents and special symbols. Might be something to reconsider in the future.

src/main/scala/org/clulab/habitus/utils/DefaultContextExtractor.scala
maxaalexeeva: So, the case class DefaultContext would not have all the strings, it would have the actual types appropriate for the content of each field, and those would be converted to strings within printers?

src/main/scala/org/clulab/habitus/utils/DefaultContextExtractor.scala
maxaalexeeva: Should I file an issue for now to investigate later? To be honest, I am not sure how this should be addressed :(

src/main/scala/org/clulab/habitus/utils/DefaultContextExtractor.scala
kwalcock: This printer might have a policy of converting None to "N/A".  I think you were using some kind of reflection, so this idea might not play nicely with that.  It's much more important to get the linguistics right than these little things, though.

python/processing_geonames/get_essential_region_info_from_geonames.py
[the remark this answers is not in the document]
kwalcock: If there are special symbols anywhere, the second (or third or fourth) byte of a UTF character might look like a tab or comma or quote or something that will mess up the parsing.  I'd see if that pd.read_csv (and any later write) has some way to specify the encoding. and google thinks that it does.

src/main/scala/org/clulab/habitus/utils/DefaultContextExtractor.scala
kwalcock: That sounds good.  If you eventually describe the situation a little more, about when these regions are being processed, it would help.  It looks like it's reading a directory, but resources don't work quite like that.  Are we expecting users to edit the configuration file that is contained inside the jar file in order to specify the directory?  Things like that.

maxaalexeeva: I think I was thinking that that place in the resources is where all regions file would go, like lexicons do, and users would not be changing that location. 
I was hoping the regions files would be read once in that default context object and then we would just use the created map. 

> but resources don't work quite like that
Would you please say more on this?

src/main/scala/org/clulab/habitus/HabitusReader.scala
```diff
@@ -53,4 +57,12 @@ class HabitusReader() extends App {
       }
     }
   }
+  def getYear(filename: String, metaFiles: Seq[File]): Option[Int] = {
+    implicit val formats = DefaultFormats
+    val file = metaFiles.filter(_.getName.contains(filename.replace(".txt", ".json")))
```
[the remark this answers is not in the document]
maxaalexeeva: Adjusted this.

python/processing_geonames/get_essential_region_info_from_geonames.py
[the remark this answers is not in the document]
maxaalexeeva: added encoding

src/main/scala/org/clulab/habitus/HabitusReader.scala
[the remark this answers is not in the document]
kwalcock: It's almost there.  I *think* that either `val metaFiles` can be removed and somewhere else a check included for `if (metaDir.isDefined)`, or metaFiles stays one one checks whether the metaFile is within the metaFiles.  I'm not sure how comparison of two Files works and would personally favor the first option myself.

kwalcock:
I was otherwise waiting for Mihai to respond, but let me know if there is a rush.

kwalcock:
@MihaiSurdeanu, I noticed you were tagged as a reviewer and know a lot more about the reason for the changes.

src/main/scala/org/clulab/habitus/HabitusReader.scala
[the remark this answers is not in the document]
maxaalexeeva: Oh right, I missed that metaFiles is not used in this way of getting the meta file. 
Adjusted more. Not sure if that's what you were looking for.

kwalcock:
That works for me.  Thanks for sticking with it.

src/main/scala/org/clulab/habitus/HabitusReader.scala
```diff
@@ -41,8 +44,9 @@ class HabitusReader() extends App {
           val unfiltered = FileUtils.getTextFromFile(file)
           val text = cleanText(unfiltered)
           val filename = StringUtils.afterLast(file.getName, '/')
```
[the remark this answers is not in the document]
MihaiSurdeanu: Not very important, but Java's File.separator should handle this:
https://www.baeldung.com/java-file-vs-file-path-separator

MihaiSurdeanu:
Very nice. Thank you @maxaalexeeva and @kwalcock!

PR #172, opened 2022-10-14
Read regions from resources

kwalcock:


kwalcock:
There's not a ready way to search for all resources of a certain name, so I suggested just recording them in a list in the configuration file.  I hope that isn't infeasible.

maxaalexeeva:
@kwalcock that looks good! Is the magic in the `sourceFromResource` method? Is it specifically for finding things within the resources directory?

maxaalexeeva (approved):

kwalcock:
@maxaalexeeva, not exactly.  At runtime there isn't really a resources directory unless you are within the development environment.  All the resources (and class files) are eventually put into the jar/zip file and special measures are needed to get them back out.  It happens with all those yml and conf files already, but opening and reading them is buried so that you don't notice.

PR #173, opened 2022-10-20
splitting planting dates into binary events

maxaalexeeva:
- there was another event that could have multiple values. That was interfering with this https://github.com/clulab/habitus/blob/36fb503755a4916ebad62fa3da958f7b72bd6dea/src/main/scala/org/clulab/habitus/utils/DefaultContextExtractor.scala#L94 and also could not be printed well in the tsv

### DIFF
--- a/src/main/resources/variables/events.yml
+++ b/src/main/resources/variables/events.yml
@@ -90,6 +90,7 @@
 - name: planting-date-cop-1
   priority: ${rulepriority}
   label: PlantingDate
+  action: splitIntoBinary
   type: token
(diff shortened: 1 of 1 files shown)


kwalcock:
Thanks, @maxaalexeeva.

kwalcock:
I *think* that `require` statement is accurate. If it's wrong, it will cause the program to crash.  Are you pretty sure there aren't other problem cases?  Something might be done in code right there so that if duplicates are encountered, they are split as needed.

PR #174, opened 2022-10-24
removing a printout

maxaalexeeva:


kwalcock:
Thanks!  In some cases procedure is just a way to keep others informed of changes.

PR #176, opened 2023-01-15
measurements fix

maxaalexeeva:
- adjusted rules to allow measurements with complex labels (incl. unit class) to account for changes in processors
- should work with both processors 8.5.2 and 8.5.3
- in future, might need to add negative constraints on some measurements (e.g., !entity='.-MEASUREMENT-<CLASS>'), but for now no tests seem to fail

### DIFF
--- a/src/main/resources/variables/entities.yml
+++ b/src/main/resources/variables/entities.yml
@@ -170,7 +170,7 @@
   type: token
   action: appropriateMeasurement
   pattern: |
-    [entity='B-MEASUREMENT' & !word = /19[0-6]\d/] [entity = 'I-MEASUREMENT' & !word = /19[0-6]\d/]*
+    [entity=/B-MEASUREMENT/ & !word = /19[0-6]\d/] [entity = /I-MEASUREMENT/ & !word = /19[0-6]\d/]*
 
 - name: measurement-rule-constructed
   priority: ${rulepriority}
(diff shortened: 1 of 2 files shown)


kwalcock:
Thanks, @maxaalexeeva.  For some reason I'm still seeing a failure inTestBeliefReader:
```
[info] - should process 0-sent1 correctly *** FAILED ***
[info]   ArrayBuffer() had size 0 instead of expected size 1 (TestBeliefReader.scala:29)
```
I notice the pattern to the changes you made and wonder whether a remaining
```
entity = 'B-FERTILIZER'
```
is intended.  It seems to fix the last test.  I'm adding it in a PR to this PR.  It's #177.

maxaalexeeva:
@kwalcock hm not sure what's going on. The test passes for me and also this test (belief reader test) does not have anything to do with the rule where the fertilizer pattern occurs (that's from the set of variable extraction rules)... Plus, 'B-FERTILIZER' comes from habitus, not processors, so it should not have been impacted. 

That said, there is no harm in having B-FERTILIZER being used as a regex, so if tests pass that way, that is fine. I just wonder if that test was just one of those randomly failing tests and it failing had nothing to do with the processors update... Am I missing something?

kwalcock:
That's an excellent point about a potentially randomly failing test!  I'll look into it.

PR #180, opened 2023-06-27
Add initial belief pipeline files

kwalcock:


kwalcock:
@maxaalexeeva, this was the pipeline so far.  I thought it should be here rather than in your repo.

maxaalexeeva:
@kwalcock makes sense! Thank you!

PR #190, opened 2023-08-23
Work on PWLWP

kwalcock:


kwalcock:
@RazvanDu, here's some space (a branch, bookmark in code) for you.  Have fun!

RazvanDu:
> @RazvanDu, here's some space (a branch, bookmark in code) for you.  Have fun!

Thanks!

pwlwp/matcher.py
```diff
@@ -8,13 +8,35 @@
...
 		self.both  =hit and is_causal and is_belief
+		self.choice_str = choice_str
+
+	def similarity(self, left_embedding, right_embedding) -> float:
+		result = numpy.dot(left_embedding, right_embedding) / numpy.linalg.norm(left_embedding) / numpy.linalg.norm(right_embedding)
+		return result
+
+class SentenceMatchFilter():
+
+	choice_str = ""
```
kwalcock, line 28: Is this a strategy to guard against uninitialized variables or to make the types more obvious or something?  There are similar sentence_matches and choice_matches later.  I'm not familiar with the practice.

pwlwp/matcher.py
```diff
@@ -67,36 +89,106 @@ def __str__(self) -> str:
...
-		]
+	def match_choice(self, introduction: str, choice: str, to_print: bool, filter_first: bool) -> ChoiceMatch:
+
+		sentence_matches = []
+
+		if not filter_first:
+
+			initial_embedding = self.sentence_transformer.encode(choice)
+			introduction_embedding = self.sentence_transformer.encode(introduction)
+			choice_embedding = numpy.add(initial_embedding, introduction_embedding) / 2
```
kwalcock, line 110: / 2 is interesting.

pwlwp/matcher.py
```diff
@@ -67,36 +89,106 @@ def __str__(self) -> str:
...
+					self.data_embeddings[index],
+					self.threshold,
+					self.threshold2,
+					is_causal=bool(self.causal_column[index]),
+					is_belief=bool(self.belief_column[index])
+				)
+				for index in range(len(self.data_frame))
+			]
+
+		if to_print:
```
kwalcock, line 143: Would this print all the same stuff with a single `if to_print:` at the beginning?

kwalcock:
Thanks, @RazvanDu.  Could you add in the meeting notes (the document in the invitation), a reminder to discuss this?  If there's something in particular you'd like to highlight, a short comment would be welcome.

pwlwp/matcher.py
RazvanDu: My code is a bit messy and I probably have to clean it, but I created 2 SentenceMatch filters depending on wether we want to first filter by the introduction or not. I could've implemented it as a Boolean variable but that would've led to more complex code. Changing to a single class is pretty easy if you want me to change it though.

pwlwp/matcher.py
RazvanDu: Yes that is true. Initially I was doing some processing alongside the printing, that's why it's structured that way. I'll fix it in a bit.

RazvanDu:
Okay, I've created a personal document which I've linked in the Habitus documentation. We can discuss it tomorrow.

pwlwp/matcher.py
kwalcock: It's just that `choice_str = ""` and a couple of similar lines that have me curious.  It doesn't seem like they are strictly necessary, so they probably serve an unfamiliar purpose.  I wouldn't change anything.  It's still research and experimental.  Maybe you can fill me in at the meeting.

pwlwp/matcher.py
kwalcock: No need yet.  There will probably be other things to change.  Thanks for checking.

pwlwp/matcher.py
RazvanDu: Oh I thought you're referring to the class as a whole. Well, "choice_str = """ is probably useless actually whops.

kwalcock (commented):
I will convert this from a draft and then probably merge it.  Then we should have @RazvanDu  make a new branch for further refinement.

I think we need to keep a (more formal, not just a screen transcript) record of at least the responses from ChatGPT in files that we can rerun locally in case we change the ranking algorithm (or find a bug, or just want to verify).  It doesn't sound like we're guaranteed to get the same answer back on a resubmission and it would also cost more.  Perhaps there would be one directory per scenario and one file per paraphrase permutation.  It would be extra nice to record the requests along with the responses.  Copies would be stored at Box.

As this code becomes more important, I'd like to see it slowly incorporate the following changes:

* Please avoid abbreviations like strr, crr, nr_paraphrases for the sake of readers.
* For simple manipulations of lists, please favor comprehensions over for statements.  I pointed a few places they can be used.
* I'd like to see some kind of Chat class that takes care of things like the name of the model, roles, unpacking the message content, etc.  It could possibly be in charge of archiving the requests and responses it sees.  It wouldn't necessarily need to rank the output though.  Shuffling the strings is sufficient.
* Type hinting should be added as other changes are made.
* We need to get a requirements file for this subproject, and openai needs to be in it.
* In addition to recording the output, I would encourage a bit more sanity checking on the format of what we get back.

Thanks, Razvan.  It's looking good.

pwlwp/cache_main.py
```diff
@@ -20,7 +20,7 @@ def doit(index, sentence):
 	return sentence_transformer.encode(sentence)
 
 if __name__ == "__main__":
-	sentence_transformer_name: str = "all-distilroberta-v1"
+	sentence_transformer_name: str = "all-MiniLM-L6-v2"
 	input_file_name: str = "../corpora/causalBeliefSentences.tsv"
```
kwalcock, line 24: I think these are still the small files.  We need to make sure this works with the large ones.

kwalcock: This includes running the sentence transformer on a million sentences.  That probably needs to be done in advance.

pwlwp/main.py
```diff
@@ -41,14 +46,129 @@ def get_in_and_out() -> str: # Tuple[str, str]:
...
+	print("CHOICES:")
+	print(choices)
+	print()
+	print("EXPLANATION:")
+	print(result)
+
+	ranks = []
+	crr = 0
+
+	for choice in choices:
```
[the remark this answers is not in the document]
RazvanDu: Yes, for the following appends.

pwlwp/main.py
```diff
@@ -41,14 +46,129 @@ def get_in_and_out() -> str: # Tuple[str, str]:
...
+		ranks.append(rank[1])
+
+	for i in range(len(ranks)):
+		ranks[i] = number_of_paraphrases - ranks[i] - 1
+
+	return ranks
+
+def compute_ranking(paraphrases, introduction, context):
+
+	final_ranks = []
```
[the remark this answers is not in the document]
RazvanDu: I've updated it.

pwlwp/main.py
```diff
@@ -41,14 +46,129 @@ def get_in_and_out() -> str: # Tuple[str, str]:
...
+		choices_chosen = []
+		for choice_list in paraphrases:
+			choices_chosen.append(choice_list[i])
+		ranks = rank_choices(introduction, context, choices_chosen)
+		for i in range(len(ranks)):
+			final_ranks[i] += ranks[i]
+
+	print(final_ranks)
+
+	sum = 0
```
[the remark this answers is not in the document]
RazvanDu: The total is the sum of the exponentials so I believe that doesn't work.

pwlwp/main.py
```diff
@@ -41,14 +46,129 @@ def get_in_and_out() -> str: # Tuple[str, str]:
...
+	for i in range(len(paraphrases)):
+		final_ranks.append(0)
+
+	for i in range(number_of_paraphrases):
+		choices_chosen = []
+		for choice_list in paraphrases:
+			choices_chosen.append(choice_list[i])
+		ranks = rank_choices(introduction, context, choices_chosen)
+		for i in range(len(ranks)):
+			final_ranks[i] += ranks[i]
```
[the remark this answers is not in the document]
RazvanDu: Yeah we can for sure explore other options.

pwlwp/main.py
```diff
@@ -41,14 +46,129 @@ def get_in_and_out() -> str: # Tuple[str, str]:
...
+	final_result = []
+
+	for strr in result:
+		index = 0
+		for i in range(len(strr)):
+			if ('a' <= strr[i] <= 'z') or ('A' <= strr[i] <= 'Z'):
+				index = i
+				break
+		if len(strr[index:]) < 2:
+			continue
```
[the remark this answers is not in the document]
RazvanDu: It seems sensitive but if I'm not mistaken I think that part has never crashed.

PR #191, opened 2023-08-24
Work on dates

kwalcock:


kwalcock:
@MihaiSurdeanu, here are how the dates are distributed.  I will check on 2017.  Perhaps those are concentrated in one source or for one search term that spiked.  I don't know that statistical tests need the same counts, but confidence would be affected.

image

MihaiSurdeanu:
Thanks @kwalcock !

This suggests we need custom queries for at least: 2014 -- 2016, 2018 -- 2020.

kwalcock:
In 2017 CITI FM had its heyday and adomonline started off enthusiastically.  ghanaweb only has articles from 2023, but that's because I specified it.  Their menu item for all years didn't seem to work.  I can probably fill in the years for that one by one.  On the others I will try to add a year to the search term and see if that helps fill in the gap.

image

PR #195, opened 2023-09-15
Work in progress on the PWLWP branch

kwalcock:


pwlwp/chat.py
```diff
@@ -0,0 +1,38 @@
...
+		self.model = model
+		self.timestr = timestr
+		os.mkdir("gpt_chat/" + timestr)
+
+	def calL_gpt(self) -> str:
+
+		chat_completion = openai.ChatCompletion.create(model=self.model,
+													   messages=[{"role": "user", "content": self.question}])
+
+		f = open("gpt_chat/" + self.timestr + "/" + str(self.current_prompt) + "_input.txt", "w")
```
kwalcock, line 26: At some point someone will go through and change these to `with` statements and check the encoding.  I see one example in the code that looks like this:
```pythons
with open(file_name, encoding="utf-8") as file:
```

kwalcock:
@RazvanDu, thanks for the updates.

pwlwp/cache_main.py
```diff
@@ -1,7 +1,5 @@
 
 from argparse import ArgumentParser
-from matcher import Matcher
```
kwalcock, line 3: Thank you!

pwlwp/matcher.py
```diff
@@ -218,20 +218,30 @@ def match_scenario(self, scenario: Scenario, to_print: bool, filter_first: bool,
...
+
 		for sentence in sentence_matches:
 			if sentence[0] < self.threshold:
 				break
-			if current_total + len(sentence[1]) <= tokens_allowed and sentence[1] not in final_matches:
+			if sentence[1] in final_matches:
+				continue
+			if current_total + len(sentence[1]) > tokens_allowed:
+				break
+			if sentence[1] not in final_matches:
```
kwalcock, line 240: Thanks for the extra break.  I think the "sentence not in" has already been taken care of above.

kwalcock:
@RazvanDu, if you aren't doing it already, you might want to use tags to mark the commits that are used for various runs.  We may need to know exactly which version was used to produce a certain output.

RazvanDu:
> @RazvanDu, if you aren't doing it already, you might want to use tags to mark the commits that are used for various runs.  We may need to know exactly which version was used to produce a certain output.

Oh, I've never used tags on GitHub, can I do that after the commits?

RazvanDu:
> @RazvanDu, if you aren't doing it already, you might want to use tags to mark the commits that are used for various runs.  We may need to know exactly which version was used to produce a certain output.

Oh, I've never used tags on GitHub, can I do that after the I pushed the commits?

kwalcock:
I usually end up doing it retroactively because I didn't think to do it at the time.  Yes, it's very possible.

kwalcock:
@RazvanDu, I usually end up doing it retroactively because I didn't think to do it at the time. Yes, it's very possible.

RazvanDu:
> @RazvanDu, I usually end up doing it retroactively because I didn't think to do it at the time. Yes, it's very possible.

Alright, I'll try to remember to do that tomorrow.

PR #203, opened 2023-09-28
Standing PR for PWLWP changes

kwalcock:


kwalcock:
There's a `print("string", file=f)` that would have been handy.  Maybe next time.

kwalcock:
If we ever need to revisit this, it would be very nice to have the results in json format.  Instead of just printing output, we can build up a json structure and write that to a file instead.  It would help with the parsing of the output for analysis.

PR #212, opened 2023-11-22
Uganda locations

maxaalexeeva:
Please see additions to the readme for the details on how the additions were made.
In some cases, if a new location was determined to be an alternative name for an existing location, it was added to column D, which is supposed to contain alternative names. I think I intended to use that column to make sure that we don't miss alternative names, but I am not sure, so it may not be used in the new version of the pipeline, @kwalcock


kwalcock:
The large diff in the UG.tsv file is primarily due to something changing 00 to 0 and then adding an extra often or always blank column at the right, plus a free conversion of -0.00007 to -7E-05.  Other than that, only 9 rows were changed and then 57 were added to the bottom.  I'm not sure it matters, but here it is for the record.

kwalcock:
@maxaalexeeva, as far as I can tell, the alternate/alternative names are still used: https://github.com/clulab/habitus/blob/54b6f30539993d936f567787ebdb65548947fcba/belief_pipeline/tpi_location_stage.py#L29-L41

kwalcock:
@maxaalexeeva, thank you from my future self who will someday wonder what we were doing here.

kwalcock (approved):
Thank you!

PR #218, opened 2023-12-12
Demo embeddings

kwalcock:


kwalcock:
@RazvanDu, your embeddings file was serialized and sent to artifactory.clulab.org.  This PR shows how it can be accessed, but does not yet implement a grounder.  Generally processors does not do the grounding; that needs to go in a different project, so I just put it here for now.

RazvanDu:
> @RazvanDu, your embeddings file was serialized and sent to artifactory.clulab.org.  This PR shows how it can be accessed, but does not yet implement a grounder.  Generally processors does not do the grounding; that needs to go in a different project, so I just put it here for now.

Cool, got it.

src/main/scala/org/clulab/habitus/apps/HabitusGrounderApp.scala
```diff
@@ -0,0 +1,50 @@
...
+      println(s"$key\t$similarity")
+      similarity
+    }
+
+    grounding
+  }
+}
+
+object HabitusGrounder {
+  // TODO: A simple "deberta-base" fails!  With v3, the special character is incorrect.
```
kwalcock, line 28: @RazvanDu, for your embedding_generator/main.py in processors, did you literally use the "microsoft/deberta-base" that is in the [comments](https://github.com/clulab/processors/blob/7f01cd92c43b5aea48963cb6b8510543a99e99f5/main/src/main/python/embedding_generator/main.py#L21) or something else when producing your deberta-embd.tsv?  The deberta-base does not seem to be available at the rust level.

RazvanDu: That weird, yes, I used the exact values in the comments/example commands and they work just fine.

kwalcock: There are some tokenizers that are only directly available to Python, but there is some way to export them.  I just have to remember how that is done.  Thanks!

RazvanDu: Ah that makes sense, gotcha.

PR #266, opened 2025-02-11
Syria Downloader Classes

arunreddy-29:


kwalcock:
@arunreddy-29, it's great that someone is using this.  I'm not sure that this is the best place for the additions.  They might be more appropriate in a branch or a fork.  Some of the files I have previously checked in only as a record of what that was worked on as part of the habitus project/grant.  It's not clear to me what you are working on.  Could you introduce yourself and let me know what this is about?  If this isn't a good place to do so, could you find my email address (check staff at http://cs.arizona.edu) and send me a message?  Thank you!

